PR #52, opened 2024-01-25
Documentation automation

evan-masseau:
I didn't think to test the fetch command locally before. The shallow fetch causes the branch to have no history, so it cannot be merged with any other branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,7 +186,6 @@ jobs:
 
       - name: Create Updated Documentation PR
         run: |
-          git fetch --depth=1 --no-tags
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -186,7 +186,6 @@ jobs:
 
       - name: Create Updated Documentation PR
         run: |
-          git fetch --depth=1 --no-tags
```
kennyklaviyo, line 189: yeah i wanted to optimize since fetching without this gets everything

evan-masseau: yeah, i dont think we need to fetch at all though, since it'll checkout documentation from origin, and i can merge straight from master

PR #68, opened 2024-01-30
Add template local.properties file for the example app

evan-masseau:
# Description

A nice way to set company ID from local properties for android. This is minor because it only applies to the example app. 

As I write this I realize it'd be nice to do it for iOS example too, though not sure how. 

# Check List

- [x] Are you changing anything with the public API? NO
- [x] Are your changes backwards compatible with previous SDK Versions? YES

### DIFF
--- a/example/android/local.properties
+++ b/example/android/local.properties
@@ -1,0 +1,5 @@
+## This file must *NOT* be checked into Version Control Systems,
+# as it contains information specific to your local configuration,
+# such as your Klaviyo company ID aka "Public API Key".
+
+#publicApiKey=YOUR_PUBLIC_API_KEY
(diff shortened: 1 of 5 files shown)


example/android/local.properties
```diff
@@ -0,0 +1,5 @@
+## This file must *NOT* be checked into Version Control Systems,
```
ajaysubra, line 1: Should you check this in?

ajaysubra (approved):
one question but looks good

example/android/local.properties
evan-masseau: oh i see what you mean, i'll change that wording

PR #70, opened 2024-01-30
Pod install | static linking issue | RCT Folly | Turbo modules

ajaysubra:
# Description

We were notified of a customer facing issues when trying to `pod install` in the iOS directory of their react native project. The issue read - 

```
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `klaviyo-react-native-sdk` depends upon `RCT-Folly.common`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

Although, we were unable to reproduce this error in our test app even after updating our podfile to match the customers, we tried to dig into the issue and it turns out that `RCT-Folly.` is an objective c library that was getting pulled in as one of our dependency because we were trying to get turbo modules to work. 

As a solution, 
* since we are not currently using turbo modules, I've updated our podspec to not include that as a dependency 
* Further, just as a safely check I've added a line to our podspec that would allow swift and objective c interoperability if in the future we ended up pulling in objective c libraries.

# Check List

## Test Plan

Once I get this PR in I'll use the `npm install` with a commit and test it locally and make sure our internal test app is able to pull in dependencies with cocoapods and works as expected. I will test it with a few version of react native to be sure that this works on all versions.


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "klaviyo-react-native-sdk",
-  "version": "0.1.0",
```
ajaysubra, line 3: This is a mistake doesn't have to be in here. I'll fix it.

klaviyo-react-native-sdk.podspec
```diff
@@ -1,10 +1,9 @@
 require "json"
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
-folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
 
 Pod::Spec.new do |s|
-  s.name         = "klaviyo-react-native-sdk"
+  s.name         =  package["name"]
```
ajaysubra, line 6: using the name from the package just to avoid having it in two places. Not related to the fix.

klaviyo-react-native-sdk.podspec
```diff
@@ -13,30 +12,10 @@ Pod::Spec.new do |s|
 
   s.platforms    = { :ios => "13.0" }
   s.source       = { :git => "https://github.com/klaviyo/klaviyo-react-native-sdk.git", :tag => "#{s.version}" }
-
   s.source_files = "ios/**/*.{h,m,mm,swift}"
 
-  # Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
```
ajaysubra, line 19: turbo module/new arch related stuff

klaviyo-react-native-sdk.podspec
```diff
@@ -13,30 +12,10 @@ Pod::Spec.new do |s|
...
-  # See https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79.
-  if respond_to?(:install_modules_dependencies, true)
-    install_modules_dependencies(s)
-  else
-    s.dependency "React-Core"
+  s.pod_target_xcconfig    = { "DEFINES_MODULE" => "YES" }
 
-    # Don't install the dependencies when we run `pod install` in the old architecture.
```
ajaysubra, line 26: turbo module/new arch related stuff

evan-masseau:
> Once I get this PR in I'll use the npm install with a commit 
You should be able to npm install with a commit hash without merging to master

ajaysubra:
> > Once I get this PR in I'll use the npm install with a commit
> 
> You should be able to npm install with a commit hash without merging to master

Yeah you are right, I did do some vague testing locally already so I was thinking of doing a more formal test once this was in but I guess you are right. Better to do the formal testing before merging so that I can fix any issue right here.

ajaysubra:
Testing on the below RN versions and posting the results of `pod install`. Please note that this doesn't indicate that we support all of these versions at the moment but mostly indicating that pod install works on these versions.

| React native version | `pod install` Result |
|-----------------|-----------------|
| 0.68.7 | ✅ |
| 0.69.12 | ✅ |
| 0.70.15 | ✅ |
| 0.71.15 | ✅ |
| 0.72.10 | ✅ |
| 0.73.3 | ✅ |

evan-masseau (approved):
Still don't totally understand why it broke for them, but If it worked for them, it works for me.
Make sure the squashed commit message clearly mentions RCT Folly, New Arch and Turbo Modules so that if we need to revert/change this in the future when we work on new arch/turbo its easy to find please!

lightstrike:
I tried running the example RN project in this repo from the `master` branch and hit a bunch of different errors with running `yarn ios` (`yarn android` was fine). Pulling down the changes from this branch resolved all the pod issues, nice work! 🎊  

Let me know if you want me to run anything from my end to try hunting down what's driving the breakage in master.

evan-masseau:
Thanks for chiming in @lightstrike, if you wouldn't mind sharing your `npx react-native info` and any of the error messages you hit that may help us understand more going forward. In our environments, the example app builds on `master`, so we've been having trouble reproducing. But we are glad to hear that this branch fixed it for you! We'll keep you posted when this gets into an official release.

lightstrike:
Sure thing @evan-masseau, here's the output of `npx react-native info` when running the example on the `master` branch:
```
System:
  OS: macOS 14.2.1
  CPU: (10) arm64 Apple M1 Max
  Memory: 85.81 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 18.4.0
    path: ~/.nvm/versions/node/v18.4.0/bin/node
  Yarn:
    version: 3.6.1
    path: ~/.nvm/versions/node/v18.4.0/bin/yarn
  npm:
    version: 8.12.1
    path: ~/.nvm/versions/node/v18.4.0/bin/npm
  Watchman: Not Found
Managers:
  CocoaPods:
    version: 1.14.3
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.2
      - iOS 17.2
      - macOS 14.2
      - tvOS 17.2
      - watchOS 10.2
  Android SDK: Not Found
IDEs:
  Android Studio: 2023.1 AI-231.9392.1.2311.11255304
  Xcode:
    version: 15.1/15C65
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.9
    path: /opt/homebrew/opt/openjdk@17/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.73.1
    wanted: 0.73.1
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false

info React Native v0.73.3 is now available (your project is running on v0.73.1).
info Changelog: https://github.com/facebook/react-native/releases/tag/v0.73.3
info Diff: https://react-native-community.github.io/upgrade-helper/?from=0.73.3
info For more info, check out "https://reactnative.dev/docs/upgrading?os=macos".
```

[Here's the current error](https://gist.github.com/lightstrike/df581d00eeb9268cd4310328fbe2c0a1) thrown by `yarn ios` (linked as it's a lot, hopefully helpful!). Let me know if you want me to check anything else.

evan-masseau:
Thanks @lightstrike, ultimately the error seems to be from `RCTThirdPartyFabricComponentsProvider.mm` which I think is part of what Ajay is taking out in this PR. It does appear to be related to the issue that prompted us to draft this PR in the first place. 
Unfortunately I still can't reproduce it myself despite the fact that according to the output from our `npx react-native info` being basically identical. Here are the steps I took with a clean clone of the repository:

- clean clone of the repo
- `yarn install` in project roo
- `cd example/ios`
- `pod install` 
- - Note, using `yarn pod-install` created an erroneous `xcode.env.local` file with a bad path to node binary, so I opt for using `pod install`.
- - Also note that this pod install modified the pod lockfile to bump to v 0.1.1 of the Klaviyo SDK, something we should update in version control
- `yarn ios` -- takes a while to build project, but succeeds. 

We certainly aren't giving up on understanding the issue, but I think this PR will have to be our first step. We're just doing a little more testing before we merge and release this. 

Thanks again for your active help. If there were any differences in how you installed and tested the example, let me know. There are often several ways to do the same things with RN, so a minor difference might give us a clue!

evan-masseau:
The discovery about `yarn pod-install` generating an erroneous local xcode config file has finally allowed me to fix our Github Action that compiles the example app on iOS, see #75. That will be a nice sanity check going forward that in a clean environment our example app does in fact compile and launch on both platforms.

evan-masseau:
Fixed conflicts

PR #71, opened 2024-01-31
Android kotlin version and min SDK fixes

evan-masseau:
# Description
These take care of some android build issues on older RN versions.
- Specify kotlin version 1.8 for compat with android sdk
- Avoid using an android 24 API `foreach` method
- Revert react-native dependency to the template syntax, and put in a local.properties file so we can override it for development.

# Check List

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,7 +1,6 @@
 buildscript {
   // Buildscript is evaluated before everything else so we can't use getExtOrDefault
   def kotlin_version = rootProject.ext.has("kotlinVersion") ? rootProject.ext.get("kotlinVersion") : project.properties["KlaviyoReactNativeSdk_kotlinVersion"]
-  def kotlinVersion = kotlin_version
 
   repositories {
     google()
@@ -113,7 +112,6 @@ dependencies {
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   api "com.github.klaviyo.klaviyo-android-sdk:analytics:2.0.0"
   api "com.github.klaviyo.klaviyo-android-sdk:push-fcm:2.0.0"
-  implementation "org.jetbrains.kotlin:kotlin-reflect:1.8.21"
 }
 
 if (isNewArchitectureEnabled()) {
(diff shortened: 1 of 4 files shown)


android/build.gradle
```diff
@@ -108,11 +108,10 @@ dependencies {
   // For < 0.71, this will be from the local maven repo
   // For > 0.71, this will be replaced by `com.facebook.react:react-android:$version` by react gradle plugin
   //noinspection GradleDynamicVersion
-  implementation "com.facebook.react:react-android:0.73.1"
+  implementation "com.facebook.react:react-native:+"
```
evan-masseau, line 111: This fixed the issue with test apps on `<=0.70`

android/gradle.properties
```diff
@@ -12,7 +12,7 @@ org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryErro
 # org.gradle.parallel=true
 #Tue Dec 19 15:08:27 EST 2023
 KlaviyoReactNativeSdk_compileSdkVersion=31
-KlaviyoReactNativeSdk_kotlinVersion=1.7.0
+KlaviyoReactNativeSdk_kotlinVersion=1.8.0
```
evan-masseau, line 15: 1.8 is compatible with 1.9.22 so we can stop here for now

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
```diff
@@ -49,7 +49,7 @@ class KlaviyoReactNativeSdkModule internal constructor(private val context: Reac
     override fun setProfile(profile: ReadableMap) {
       val parsedProfile = Profile()
 
-      profile.toHashMap().forEach { (key, value) ->
```
evan-masseau, line 52: This triggers a warning about min SDK version of 24.

android/build.gradle
evan-masseau: It causes kotlin files to appear broken while editing, unfortunately, because facebook/react imports appear broken. don't know how to fix that yet, but it compiles from the example app.

kennyklaviyo (approved):

ajaysubra (approved):

PR #75, opened 2024-02-05
Github CI Updates

evan-masseau:
[Coincidentally discovered](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/70#issuecomment-1928101370) that `yarn pod-install` seems to create an erroneous `xcode.env.local` file pointing to a non-existent node binary. I recalled that the hermes error that was causing iOS CI step to fail is caused by a bad node binary. I had tried to fix it before by adding my own local config file, but I think `yarn pod-install` was overwriting it 🙃 

Also moved the doc bot to its own file so it doesn't have to appear as a skipped step on all pull requests. 

#70


evan-masseau:
FYI I tried using latest version of `pod-install` package too, still has the same problem. See the test runs for #79

ajaysubra (approved):

PR #84, opened 2024-02-06
Release 0.1.2 

ajaysubra:
# Description

# Check List


evan-masseau (changes requested):
i think you need to run pod install in the example app to update its pod lockfile

ajaysubra:
> i think you need to run pod install in the example app to update its pod lockfile

Thanks for reminding. Done.

evan-masseau (approved):

PR #86, opened 2024-02-06
Break up CI files so we can get individual badges

evan-masseau:
Adds badges for CI to the readme, had to split up the yml files so i could have a badge per platform.
In the release automation PR i will change these so that they are linked to _release_ rather than just the latest merge to master.

### DIFF
--- a/.github/workflows/ios-ci.yml
+++ b/.github/workflows/ios-ci.yml
@@ -1,0 +1,68 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+permissions:
+  contents: write
+
+jobs:
+  build-ios:
+    runs-on: macos-13-xlarge
+    env:
+      TURBO_CACHE_DIR: .turbo/ios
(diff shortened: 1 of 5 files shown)


.github/workflows/ios-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+          restore-keys: |
+            ${{ runner.os }}-turborepo-ios-
+
+      - name: Check turborepo cache for iOS
+        run: |
+          TURBO_CACHE_STATUS=$(node -p "($(yarn turbo run build:ios --cache-dir="${{ env.TURBO_CACHE_DIR }}" --dry=json)).tasks.find(t => t.task === 'build:ios').cache.status")
+
+          if [[ $TURBO_CACHE_STATUS == "HIT" ]]; then
+            echo "turbo_cache_hit=1" >> $GITHUB_ENV
+          fi
```
ajaysubra, line 45: do we need these two steps?

ajaysubra (approved):

.github/workflows/ios-ci.yml
evan-masseau: I think its some mechanism to speed up repeated executions by caching dependencies.

PR #97, opened 2024-02-09
[WIP] Expanding on example app instructions

evan-masseau:
DRAFT 
From conversation with @ajaysubra about the gemfile and the example app. I wanted to commit some thoughts before I lose them, but I should actually ticket this out for discussion/speccing.

The idea here is to use `bundle` to define dependencies better (e.g. cocoapods and ruby version), and then of course include instructions about how to run the example app.


.github/workflows/ios-ci.yml
```diff
@@ -58,8 +58,8 @@ jobs:
       - name: Install cocoapods
         if: env.turbo_cache_hit != 1 && steps.cocoapods-cache.outputs.cache-hit != 'true'
         run: |
-          cd example/ios
-          pod install
+          yarn example-setup
+          rm -f example/ios/.xcode.env.local
```
evan-masseau, line 62: even bundle pod install causes the node_binary path to break

ajaysubra (approved):

PR #99, opened 2024-02-09
Updated iOS instructions in example app

ajaysubra:
# Description

Added, 
1. Foreground notification handling
2. Background notification handling 
3. Open tracking 
4. Deep links using URL scheme

to the example app.

# Check List


example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
```diff
@@ -4,55 +4,97 @@
...
 
+// Installation Step 8: [Optional] Implement this if registering with APNs fails
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error {
+  if (isDebug) {
     NSLog(@"Failed to register for remote notifications: %@", error);
+  }
 }
 
-// Handle push notifications when the app is in the background or terminated
+// Installation Step 9: Implement the delegate didReceiveNotificationResponse to handle push notifications when the app is in the background
```
evan-masseau, line 47: Clarifying question on this comment

> to handle push notifications when the app is in the background

This delegate method handles the user response to a notification right? like tapping on it?

ajaysubra: Yeah, it does. Were you thinking we should include that in as well? I do agree that the word "handle" is doing a lot of heavy lifting here.

ajaysubra: Updated, let me know.

evan-masseau (approved):

PR #104, opened 2024-02-14
Implemented deep links to call react native in example iOS app

ajaysubra:
# Description

Currently deep links were handled on the iOS side. Added some logic to call into react native listeners.

# Check List


example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
```diff
@@ -90,12 +92,11 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center
 
 // Installation Step 13: Implement this method to receive deep link. There are some addition setup steps needed as mentioned in the readme here -
 // https://github.com/klaviyo/klaviyo-swift-sdk?tab=readme-ov-file#deep-linking
-// additionally routing to the right screen in the app based on the url is also something that should be handled
+// Calling `RCTLinkingManager` is required for your react native listeners to be called
 - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
-  return [PushNotificationsHelper handleDeepLinksWithUrl:url];
```
ndurell, line 95: What did this do?

ndurell: ahh

example/src/App.tsx
```diff
@@ -1,10 +1,22 @@
...
+
 import { Button, View } from 'react-native';
 import { type AppViewInterface, appViews } from './AppViewInterface';
 import { styles } from './Styles';
+import { Linking } from 'react-native';
 
 export default function App() {
+  useEffect(() => {
+    Linking.addEventListener('url', ({ url }) => {
+      console.log('Event Listener: url', url);
```
ndurell, line 13: I like this for now - maybe in the future we could add some kind of navigation.

ndurell (approved):

example/src/App.tsx
ndurell: Might be tricky in the example app though.

example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
ajaysubra: I was just handling like we'd do on iOS. Now it will route to RN where customer can decide how to navigate to their pages.

example/src/App.tsx
ajaysubra: Might be possible since it's just navigating within RN. Let me see if I can find a quick example. Good idea though.

ajaysubra: Going to leave it as is for now and cut out a task to handle this later as it looks a little more involved.

ndurell: Yah that makes more sense to me. This is good for now!

evan-masseau (approved):

ajaysubra:
@ndurell @evan-masseau added calling RCTLinkHandler from the deep linking completion handler so that folks who want to intercept links/urls can do so without going through the system and it calling the application open delegate.

PR #105, opened 2024-02-15
Update README.md

ajaysubra:
# Description

Fixed the ordering of deep links (URL scheme and universal links) on the react native readme.

# Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ instructions below.
   With option 2, when you get the `deepLinkHandler`, you can handle it as follows:
 
   ```objective-c
-  [RCTLinkingManager application:application openURL:url options:options]
+    [RCTLinkingManager application:UIApplication.sharedApplication openURL: url options: @{}];
   ```
 
   For application, you can pass in an instance of `UIApplication` and since you won't have `options`, you can just pass in an empty dictionary for that parameter.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -255,16 +255,17 @@ instructions below.
...
 
   ```objective-c
   [RCTLinkingManager application:application openURL:url options:options]
   ```
 
-  Since you won't have `options`, you can just pass in an empty dictionary for that parameter.
-
-  With option 2, when you handle the open url (in [`application(_:open:options)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623112-application)),
-  you call the linking code block above similar to what you would do with option 1.
+  For application, you can pass in an instance of `UIApplication` and since you won't have `options`, you can just pass in an empty dictionary for that parameter.
```
ndurell, line 268: should we just update the code too?

ajaysubra: Yup. doing that in the example app PR - https://github.com/klaviyo/klaviyo-react-native-sdk/pull/104

ajaysubra: Oh you mean the sample code in the readme?

ndurell: Yah just worried someone would copy/paste it wrong. But they could probably figure it out.

ndurell (approved):

README.md
ajaysubra: added code, let me know if that's what you meant.

PR #116, opened 2024-02-21
Update to version 0.2.0

evan-masseau:
Version bump, and deleted npm lock file since we aren't using it and it keeps confusing me.


ajaysubra (approved):
@evan-masseau any reason for this to still be in draft?

PR #120, opened 2024-02-26
Ecm/ci cache skip

evan-masseau:


example/ios/Podfile.lock
```diff
@@ -1383,4 +1383,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 4b9faf5e512fab6e8d87beb30570ec4d3bd884b9
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.14.3
```
ajaysubra, line 1386: Should we update the gemfile to use the later version?

evan-masseau: I keep forgetting 1.15.2 is fixed! Let me just check if it's possible to use 1.15.2 from the GitHub box

ajaysubra: def not a blocking comment. WIll approve.

ajaysubra (approved):

example/ios/Podfile.lock
evan-masseau: Actually, given the obnoxious compatibility issues we've seen so far, I'm inclined to just stick to the cocoa pods version that this react native version was built with. We can update cocoa pods when we next update the react native version itself id say

Issue #129, opened 2024-03-04
'klaviyo_react_native_sdk-Swift.h' file not found


niklavsBariss:
### Checklist

- [X] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla project, such as the example app in this repository.
- [x] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [X] This issue hasn't been addressed in an [existing issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls)

### Description

```markdown
Using version klaviyo-react-native-sdk 0.2.0 and React Native 0.71.13

After a successful `pod install` trying to launch an app on Xcode(15.2) simulator results in: 

../node_modules/klaviyo-react-native-sdk/ios/KlaviyoReactNativeSdk.mm:2:9 'klaviyo_react_native_sdk-Swift.h' file not found
```

### Steps to reproduce

1. `yarn add klaviyo-react-native-sdk`

2. Add code to Podfile
```
MIN_IOS_OVERRIDE = '13.0'
if Gem::Version.new(MIN_IOS_OVERRIDE) > Gem::Version.new(min_ios_version_supported)
    min_ios_version_supported = MIN_IOS_OVERRIDE
end
```
3. `pod install`
4. Run from XCode _or_ `yarn ios`

I tried doing the same with a clean react native 0.71.13 project, but didn't have the same issue.

### The Klaviyo React Native SDK version information

0.2.0

### Environment Description
```
OS: macOS 14.3
CPU: (10) x64 Apple M1 Pro
Node: 18.19.0
Yarn: 1.22.17
CocoaPods: 1.15.2
Xcode: 15.2/15C500b
react: 18.2.0
react-native: 0.71.13

ajaysubra:
[...]

Issue #94, opened 2024-02-08
'klaviyo_react_native_sdk-Swift.h' file not found


yasicmd:
### Description

```markdown
When I run the project using Xcode, I get the following error

'klaviyo_react_native_sdk-Swift.h' file not found

file path /Users/yasi/Appmaker/staging/node_modules/klaviyo-react-native-sdk/ios/KlaviyoReactNativeSdk.mm

I did the pod install and it shows in the Podfile.lock as below

- klaviyo-react-native-sdk (0.1.2):
    - KlaviyoSwift (= 3.0.2)
    - React-Core
- KlaviyoSwift (3.0.2):
    - AnyCodable-FlightSchool

Note
----
This issue happens in both my project and in the example project provided no error showned.


### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla project
- [ ] If possible, I've reproduced the issue using the `master` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls).

### Expected behavior

```markdown
It should build the project after the pod install has been successfully done.
```

### Actual behavior

```markdown
'klaviyo_react_native_sdk-Swift.h' file not found
```

### The Klaviyo React Native SDK version information

0.1.2

### Environment Description

```
System:
  OS: macOS 14.1
  CPU: (8) arm64 Apple M1
  Memory: 92.23 MB / 8.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 18.16.0
    path: ~/.nvm/versions/node/v18.16.0/bin/node
  Yarn:
    version: 1.22.21
    path: ~/.nvm/versions/node/v18.16.0/bin/yarn
  npm:
    version: 9.5.1
    path: ~/.nvm/versions/node/v18.16.0/bin/npm
  Watchman:
    version: 2024.01.22.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.15.2
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.0
      - iOS 17.0
      - macOS 14.0
      - tvOS 17.0
      - watchOS 10.0
  Android SDK: Not Found
IDEs:
[...]

PR #78, opened 2024-02-06
Fix import of Klaviyo Swift SDK

youcefswoo:
# Description

Fix the compile issue with XCode.

image

# Check List

- [x] Are you changing anything with the public API? No
- [x] Are your changes backwards compatible with previous SDK Versions? Yes

## Test Plan

Trying to build using XCode.

PR #132, opened 2024-03-07
Fixed Xcode-generated header file import

ajaysubra:
# Description

Fixed the issue where certain developers who were importing our SDK as a framework (over cocoapods default of static library) were facing issues with the Xcode-generated header file. 

This was happening mainly for folks using firebase as it instructs developers to add a line to you pod file that in turn updates all the pods to be pulled in as frameworks.  On inspection we found [this](https://developer.apple.com/documentation/swift/importing-swift-into-objective-c) apple doc that suggests using different header import syntax when importing a framework. It doesn't talk about libraries but we think that the syntax when our SDK is pulled in as a library (static or dynamic) is different. 

As a fix, we included importing both formats to accommodate users pulling in our SDK as a framework and library by adding a conditional . The solution is similar to what was proposed in #78 but includes support for library import.

# Check List

- Are you changing anything with the public API? ⛔  
- Are your changes backwards compatible with previous SDK Versions?  ⛔ 

## Changelog / Code Overview

Just a simple change but took forever to reproduce internally since our test apps were not using frameworks and importing our SDK as a library.

## Test Plan

* Tested in - example app in this repo, internal test app and also tested on various older versions of react native using an internal automated script.

## Related Issues/Tickets

https://github.com/klaviyo/klaviyo-react-native-sdk/issues/94
https://github.com/klaviyo/klaviyo-react-native-sdk/pull/78/files
https://github.com/klaviyo/klaviyo-react-native-sdk/issues/129


ios/KlaviyoReactNativeSdk.mm
```diff
@@ -1,5 +1,10 @@
 #import "KlaviyoReactNativeSdk.h"
-#import <klaviyo_react_native_sdk-Swift.h>
+#if __has_include(<klaviyo_react_native_sdk/klaviyo_react_native_sdk-Swift.h>)
```
evan-masseau, line 2: nice

evan-masseau (approved):

example/ios/Podfile
```diff
@@ -28,13 +28,15 @@ end
...
+  use_frameworks! :linkage => :static
+
   use_react_native!(
     :path => config[:reactNativePath],
     # Enables Flipper.
     #
     # Note that if you have use_frameworks! enabled, Flipper will not work and
     # you should disable the next line.
-    :flipper_configuration => flipper_config,
+#    :flipper_configuration => flipper_config,
```
kennyklaviyo, line 39: was this intentional?

kennyklaviyo: nvm i see the comment

kennyklaviyo (approved):

PR #135, opened 2024-03-13
CHNL-6305: Added set profile attributes bridge method

ajaysubra:
# Description

During the initial release of our beta react native SDK, we somehow missed bridging methods for `setProfileAttributes` on iOS. This PR aims to add those missing bridging methods. 

# Check List

## Test Plan

Performed manual testing using different profile property attributes such as latitude, longitude, city etc.


example/ios/Podfile.lock
```diff
@@ -1379,7 +1379,7 @@ SPEC CHECKSUMS:
   React-utils: debda2c206770ee2785bdebb7f16d8db9f18838a
   ReactCommon: ddb128564dcbfa0287d3d1a2d10f8c7457c971f6
   SocketRocket: f32cd54efbe0f095c4d7594881e52619cfe80b17
-  Yoga: 2b33a7ac96c58cdaa7b810948fc6a2a76ed2d108
```
ajaysubra, line 1382: not sure why this changed 🤷🏽

example/src/KlaviyoReactWrapper.ts
```diff
@@ -97,7 +97,9 @@ export const setPushToken = async () => {
 
 export const setProfileAttribute = async () => {
   try {
-    Klaviyo.setProfileAttribute('CUSTOM', generateRandomName(12));
+    Klaviyo.setProfileAttribute(ProfileProperty.CITY, generateRandomName(5));
```
ajaysubra, line 100: added a couple properties outside of the custom one so that manual testing is easier.

evan-masseau: Nice.
Yeah I'd love to update this sample app someday to have an input form here instead of just generating random data.

ios/KlaviyoBridge.swift
```diff
@@ -82,6 +82,11 @@ public class KlaviyoBridge: NSObject {
       KlaviyoSDK().set(profile: profile)
   }
 
+  @objc
+  public static func setProfileAttribute(_ key: String, value: String) {
+      KlaviyoSDK().set(profileAttribute: getProfileKey(key), value: value)
```
evan-masseau, line 87: Doesn't really matter, but in the Android code we just used CUSTOM for all the keys 🤷 
```

    @ReactMethod
    override fun setProfileAttribute(
      propertyKey: String,
      value: String,
    ) {
      Klaviyo.setProfileAttribute(ProfileKey.CUSTOM(propertyKey), value)
    }
```

evan-masseau (approved):
LGTM

evan-masseau:
iOS CI failed because of cached cocoapods mismatching with that lockfile change. I still don't understand why it is using cache if the lockfile changed. But I set off a re-run, which should ignore cache, lets see!
Either way, I wouldn't hold up merging this due to the CI failure!

jordan-griffin (approved):

PR #145, opened 2024-04-09
Adding rich push setup for demo

ajaysubra:
# Description

Since we didn't have a demo example for rich push and a developer reported having some issues with it, adding an example here. 

# Check List

### DIFF
--- a/example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
+++ b/example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
@@ -12,7 +12,7 @@ import UserNotifications
 // MARK: notification service extension implementation.
 
 /// When push payload is marked as there being mutable-content this service
-/// (more specifically the `didReceiveNotificationRequest` ) is called to perform
+/// (more specifically `didReceiveNotificationRequest` ) is called to perform
 /// tasks such as downloading images and attaching it to the notification before it's displayed to the user.
 ///
 /// There is a limited time before which `didReceiveNotificationRequest`  needs to wrap up it's operations
--- a/example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
+++ b/example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
@@ -5,20 +5,11 @@
 //  Created by Ajay Subramanya on 4/9/24.
 //
 
+// Installation Step 15: Create this notification service extension using the steps outlined here - https://github.com/klaviyo/klaviyo-swift-sdk#Rich-Push
 import KlaviyoSwiftExtension
 import UIKit
 import UserNotifications
 
-// MARK: notification service extension implementation.
-
-/// When push payload is marked as there being mutable-content this service
-/// (more specifically `didReceiveNotificationRequest` ) is called to perform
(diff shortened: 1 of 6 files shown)


example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
+//
```
evan-masseau, line 1: In other places, you had included a comment like `Setup Step #`, thought that was nice for searchability. Would it make sense to label this file, or a particular spot, with that same comment string?

example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
...
+
+import KlaviyoSwiftExtension
+import UIKit
+import UserNotifications
+
+// MARK: notification service extension implementation.
+
+/// When push payload is marked as there being mutable-content this service
+/// (more specifically the `didReceiveNotificationRequest` ) is called to perform
```
evan-masseau, line 15: super nit
```suggestion
/// (more specifically `didReceiveNotificationRequest` ) is called to perform
```

evan-masseau (approved):
works for me! left some non-blocking comments inline, and one additional tiny suggestion since you're looking at example app:

example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
...
+import UIKit
+import UserNotifications
+
+// MARK: notification service extension implementation.
+
+/// When push payload is marked as there being mutable-content this service
+/// (more specifically `didReceiveNotificationRequest` ) is called to perform
+/// tasks such as downloading images and attaching it to the notification before it's displayed to the user.
+///
+/// There is a limited time before which `didReceiveNotificationRequest`  needs to wrap up it's operations
```
jordan-griffin, line 18: nit: `There is limited time`, the `a` in this sentence looks unnecessary

example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
```diff
@@ -0,0 +1,53 @@
...
+//  Created by Ajay Subramanya on 4/9/24.
+//
+
+import KlaviyoSwiftExtension
+import UIKit
+import UserNotifications
+
+// MARK: notification service extension implementation.
+
+/// When push payload is marked as there being mutable-content this service
```
jordan-griffin, line 14: total nit but: `When a push payload` sounds a little better

jordan-griffin: Or maybe even `When a push payload is marked as having mutable-content this service`...

ajaysubra: I removed this as this is sample code from iOS repo and probably not required here. Will fix it there in a different PR.

example/ios/KlaviyoReactNativeSdkExampleExtension/NotificationService.swift
ajaysubra: Same remove those comments.

Issue #140, opened 2024-04-02
Linking.getInitialURL() always returning null in IOS when notification open the app


youcefswoo:
### Checklist

- [X] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/.github/CONTRIBUTING.md)
- [X] I have determined whether this bug is also reproducible in a vanilla project, such as the example app in this repository.
- [X] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [X] This issue hasn't been addressed in an [existing issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls)

### Description

```markdown
After following the documentation and the example in the repository, I began testing to see if I could receive the deep link when the app is killed. In Android, it works well, but in iOS, Linking.getInitialURL() always returns null.
```

### Expected behavior

```markdown
I expected to receive the deep link that I placed in the deep link section of the notification definition.
```

### Actual behavior

```markdown
Linking.getInitialURL() always returns null in iOS when a notification is clicked.
```

### Steps to reproduce

```markdown
1. Create a notification with a deep link for IOS.
2. Click on the notification to open the application.
```

### The Klaviyo React Native SDK version information

0.3.0

### Environment Description

Using real device (iPhone 12 . IOS 16.6.1)
react native 0.73.3

ajaysubra:
@youcefswoo Couple questions - 
1. Can you confirm that you followed the [instructions](https://github.com/klaviyo/klaviyo-react-native-sdk/tree/master?tab=readme-ov-file#deep-linking) from the readme?
2. Are you able to reproduce this issue using the example app here?

When we published the example iOS app, we tested and deep linking was working but does involve some intricate setup for iOS outlined in the readme.

youcefswoo:
> @youcefswoo Couple questions -
> 
> 1. Can you confirm that you followed the [instructions](https://github.com/klaviyo/klaviyo-react-native-sdk/tree/master?tab=readme-ov-file#deep-linking) from the readme?
> 2. Are you able to reproduce this issue using the example app here?
> 
> When we published the example iOS app, we tested and deep linking was working but does involve some intricate setup for iOS outlined in the readme.

@ajaysubra 
1. Yes, I took all the time to follow the instructions and I double-checked all the steps multiple times. For the options, I used the first one since it is faster and easier.
[...]

PR #148, opened 2024-04-15
Example app | Updated app delegate to fix initial url being null

ajaysubra:
# Description

An [issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues/140) was raised by one of our developers citing that deep linking was not working when the app is terminated. On digging noticed that this is an issue with react native so added a workaround for our developers to use when they encounter this issue. Since this is not an SDK issue there is no fix at the SDK level and developers would have to add this workaround to their native iOS app. 

# Check List

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -51,3 +51,59 @@ there are two possible reasons for this:
 2. Firebase iOS SDK also swizzles AppDelegate methods when configured on your iOS app. If after disabling notifee,
    if the delegates are still not called, this may be the reason. Method swizzling can be turned off by following
    [Firebase's documentation](https://firebase.google.com/docs/cloud-messaging/ios/client).
+
+### Deep links in push notifications not getting sent over to React Native layer when the iOS app is terminated
+
+When the iOS app is terminated (killed from the app switcher by swiping up), there is a bug in React Native that 
+prevents the native layer from calling the listeners set up in React Native to listen to incoming deep links. 
+The listener in question here is `Linking.getInitialURL()` which is expected to be called when a deep link is 
+available while the app is terminated.
+
+You can find the open issue describing this problem on React Native's GitHub repository [Here](https://github.com/facebook/react-native/issues/32350). 
+There are many workarounds suggested in this issue's thread and a few other similar issues. 
+However, the workaround that worked for us involves intercepting the launch arguments in the app delegate and adding 
+a key `UIApplicationLaunchOptionsURLKey`, which React Native expects to be present when calling the `Linking.getInitialURL()` listener.
+
+
+Here's a method to implement this workaround:
+
+
+```swift 
+- (NSMutableDictionary *)getLaunchOptionsWithURL:(NSDictionary * _Nullable)launchOptions {
+  NSMutableDictionary *launchOptionsWithURL = [NSMutableDictionary dictionaryWithDictionary:launchOptions];
+  if (launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey]) {
+    NSDictionary *remoteNotification = launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey];
+    
+    if (remoteNotification[@"url"]) {
+      NSString *initialURL = remoteNotification[@"url"];
+      if (!launchOptions[UIApplicationLaunchOptionsURLKey]) {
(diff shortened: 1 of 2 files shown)


Troubleshooting.md
```diff
@@ -51,3 +51,59 @@ there are two possible reasons for this:
...
+```swift 
+- (NSMutableDictionary *)getLaunchOptionsWithURL:(NSDictionary * _Nullable)launchOptions {
+  NSMutableDictionary *launchOptionsWithURL = [NSMutableDictionary dictionaryWithDictionary:launchOptions];
+  if (launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey]) {
+    NSDictionary *remoteNotification = launchOptions[UIApplicationLaunchOptionsRemoteNotificationKey];
+    
+    if (remoteNotification[@"url"]) {
+      NSString *initialURL = remoteNotification[@"url"];
+      if (!launchOptions[UIApplicationLaunchOptionsURLKey]) {
+        launchOptionsWithURL[UIApplicationLaunchOptionsURLKey] = [NSURL URLWithString:initialURL];
```
evan-masseau, line 80: So the RN linking code expects this key to be set, but it isn't set by default by iOS when tapping a notification?

evan-masseau (approved):
The workaround makes sense to me from my limited recollection of different types of deep links. Plus, it worked for the original reporter of the issue.

Troubleshooting.md
ajaysubra: Yeah, so there are a bunch of things going on here, and I don't fully understand everything going on BTS, but here's what I know - 

1. [`getInitialUrl`](https://github.com/facebook/react-native/blob/8c53ac60782d8ccd2ccd7706d41e60cda93e2988/packages/react-native/Libraries/LinkingIOS/RCTLinkingManager.mm#L154) is the listener that is used on react native and it depends on the launch options key `UIApplicationLaunchOptionsURLKey`. This key is present when the app is invoked from a non-push deep link (like from an email/SMS) but absent when the app is launched (cold start) from a push notification in which case the launch argument contains the push payload within the key `UIApplicationLaunchOptionsRemoteNotificationKey`. Since `getInitialUrl` looks up the key `UIApplicationLaunchOptionsURLKey` this workaround moves the deep link url in the push payload to the top level of launch arguments where `getInitialUrl` will find it and the RN listeners will get the deep link value. I think this is a case where Apple being inconsistent with deep linking (push vs non-push) is an issue but the workaround I added could very well be added to the `getInitialUrl` method on RN to fix this issue.
2. The above point was when the app is terminated, but in all other cases, the other listener `Linking.addEventListener` is used and this is called whenever the app is in the foreground/background by the `RCTLinkingManager`. Why this is not used in all cases is not clear to me, if I had to take a wild guess it's got something to do with the RN run time not being initialized during cold start.

ajaysubra: @evan-masseau  Thanks for asking the question. Made me dig a little deeper.

Troubleshooting.md
```diff
@@ -51,3 +51,59 @@ there are two possible reasons for this:
 2. Firebase iOS SDK also swizzles AppDelegate methods when configured on your iOS app. If after disabling notifee,
    if the delegates are still not called, this may be the reason. Method swizzling can be turned off by following
    [Firebase's documentation](https://firebase.google.com/docs/cloud-messaging/ios/client).
+
+### Deep links in push notifications not getting sent over to React Native layer when the iOS app is terminated
+
+When the iOS app is terminated (killed from the app switcher by swiping up), there is a bug in React Native that
```
kennyklaviyo, line 57: not sure how much time itll take but perhaps itll be worth going thru a few versions of RN and call out which versions have this bug

Troubleshooting.md
evan-masseau: Awesome, that makes a lot of sense!

evan-masseau (approved):

kennyklaviyo (approved):

Troubleshooting.md
ajaysubra: Yeah I did think of that but none of the react native issue threads that I've seen so far don't seem to have that info. In the absence of that we need to manually test this on older versions to find which as you eluded is going to take a while 😢  Going to merge this in as I think what we have here is indeed helpful for folks with the issue.

PR #152, opened 2024-05-09
Updated occurrences of `String` to `string`

ajaysubra:
# Description

In TypeScript, `string` is a primitive type, while `String` is a global object that is a wrapper around the primitive `string` type. We were using both of these interchangeably which could lead to issues like this - 

```
let existingEmail = Klaviyo.getEmail(undefined);
    Klaviyo.setProfile({
      email: existingEmail, //this is an error: TS2322: Type String | null is not assignable to type string | undefined
    });
```

The above error is because `getEmail` returns an instance of type `String` whereas `setProfile` accepts an instance of `Profile` whose email property is of type `string`.

The fix was to across the board using `string` instead of `String`. This in some cases could be a breaking change if developer made any provision in their code to handle this inconsistency. Since we are in beta we will release this change as a minor version update or roll it into our first major version release. 

# Check List

### DIFF
--- a/src/Profile.ts
+++ b/src/Profile.ts
@@ -16,44 +16,44 @@ export interface KlaviyoProfileApi {
    * Update a profile's external ID.
    * @param externalId - The external ID to set
    */
-  setExternalId(externalId: String): void;
+  setExternalId(externalId: string): void;
 
   /**
    * Retrieve a profile's external ID.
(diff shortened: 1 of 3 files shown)


kennyklaviyo (approved):

evan-masseau (approved):
Thank you!

PR #153, opened 2024-05-13
Reviewed sample app with an eye to android setup instructions and opened push event issues

evan-masseau:
# Description
I'm reviewing the sample app (mostly android) to make sure it contains and explains all the setup steps, similar to what Ajay's already done for iOS. These updates should help people get set up properly to avoid problems with opened push events also. 

# Check List

- [x] Are you changing anything with the public API? -- NO
- [x] Are your changes backwards compatible with previous SDK Versions? -- Yep

## Changelog / Code Overview
- Add missing setup to the android side of sample app
- Commenting android code with all the steps
- Some updates to readme and the iOS sample code

## Test Plan
- [x] When I finish, I'll confirm the sample app works and also go back through the test app to mimic these setup steps, where they differ.

## Related Issues/Tickets
CHNL-7370


example/android/app/build.gradle
```diff
@@ -116,6 +116,7 @@ dependencies {
     // The version of react-native is set by the React Native Gradle Plugin
     implementation("com.facebook.react:react-android")
     implementation("com.facebook.react:flipper-integration")
+    implementation 'com.google.firebase:firebase-messaging-ktx:24.0.0'
```
ndurell, line 119: huh is this not a transitive dependency?

ndurell: Do we have this instruction in our readme or is it just something we are using for this test app?

evan-masseau: Nope, I do wonder if it should be.

ndurell: Should you add a comment about why this is needed?

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainActivity.kt
```diff
@@ -17,4 +32,113 @@ class MainActivity : ReactActivity() {
    * which allows you to enable New Architecture with a single boolean flags [fabricEnabled]
    */
   override fun createReactActivityDelegate(): ReactActivityDelegate = DefaultReactActivityDelegate(this, mainComponentName, fabricEnabled)
+
+  /**
+   * Launches a permission request, and receives the result in the callback below
+   */
+  private var requestPermissionLauncher: ActivityResultLauncher<String> =
```
ndurell, line 39: There's no way in RN to do this right?

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainApplication.kt
```diff
@@ -41,15 +39,27 @@ class MainApplication : Application(), ReactApplication {
...
-      Klaviyo.initialize(BuildConfig.PUBLIC_API_KEY, this)
-    }
-
     if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
       // If you opted-in for the New Architecture, we load the native entry point for this app.
       load()
     }
     ReactNativeFlipper.initializeFlipper(this, reactNativeHost.reactInstanceManager)
+
+    if (USE_NATIVE_IMPLEMENTATION) {
```
ndurell, line 48: Oh neat idea!

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainActivity.kt
evan-masseau: Yeah there is, there are 3rd party libs for cross platform permissions. I think right now we mention them in the readme but don't implement them in the sample app

example/android/app/build.gradle
evan-masseau: Good call bc i already forget why I had to add it

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainActivity.kt
ndurell: Oh I thought I saw something in the .74 release but it was unrelated to push.

evan-masseau: this is just a generic permission request launcher, works roughly the same regardless of permission type (with some exceptions, like geolocation permissions have additional tiers)

example/src/KlaviyoReactWrapper.ts
```diff
@@ -50,7 +52,7 @@ export const setExternalId = async () => {
 export const getEmail = async () => {
   try {
     Klaviyo.getEmail((value: string) => {
-      console.log('email is =', value);
+      console.log('email = ', value);
```
ndurell, line 55: Is there a way to do a toast in rn?

evan-masseau: It'd be a nice addition yeah, I think in first pass we were keeping this sample app light on dependencies, but I imagine there's a cross-platform toast lib. iOS doesn't have toast exactly, but you can use alerts in a less intrusive way.

ndurell (approved):

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainApplication.kt
```diff
@@ -41,15 +39,27 @@ class MainApplication : Application(), ReactApplication {
...
+      // Android Installation Step 3: Initialize the SDK with public key and context, if initializing from native code
+      Klaviyo.initialize(BuildConfig.PUBLIC_API_KEY, this)
+
+      // Android Installation Step 4a: Collect push token and pass it to Klaviyo, if handling push tokens natively
+      FirebaseMessaging.getInstance().token.addOnSuccessListener {
```
evan-masseau, line 53: Oh I think this right here is why I needed the firebase dependency @ndurell

example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
```diff
@@ -10,38 +10,38 @@ @implementation AppDelegate
 BOOL isDebug = YES;
 
 // Change to NO if you prefer to initialize and handle push tokens in the React Native layer
-BOOL useNativeImplementation = YES;
```
ajaysubra, line 13: Intentional? If yes probably should update the comment right above.

ajaysubra (approved):
1 minor comment else LGTM.

example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
evan-masseau: definitely not intentional, i think i have a couple test things in this PR that need to switched back!

example/android/app/build.gradle
```diff
@@ -74,15 +74,21 @@ android {
...
     ndkVersion rootProject.ext.ndkVersion
     buildToolsVersion rootProject.ext.buildToolsVersion
     compileSdk rootProject.ext.compileSdkVersion
 
     namespace "com.klaviyoreactnativesdkexample"
     defaultConfig {
-        applicationId "com.klaviyoreactnativesdkexample"
+        // Optional Android Installation Step: Set the applicationId if you want to run this sample app with
+        // firebase enabled. It must match an applicationId in your google-services.json file
+        applicationId "com.klaviyorntestapp"
```
evan-masseau, line 91: Mistake!

example/android/app/build.gradle
evan-masseau: Unresolved: didn't actually add a comment here!

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainActivity.kt
```diff
@@ -17,4 +33,139 @@ class MainActivity : ReactActivity() {
    * which allows you to enable New Architecture with a single boolean flags [fabricEnabled]
    */
   override fun createReactActivityDelegate(): ReactActivityDelegate = DefaultReactActivityDelegate(this, mainComponentName, fabricEnabled)
+
+  /**
+   * Launches a permission request, and receives the result in the callback below
```
evan-masseau, line 38: Add comment that this is optional for native, and can be done from RN as alternative?

PR #161, opened 2024-06-07
Fix phone number attribute on example

giorgiofellipe:
# Description

# Check List


ajaysubra (approved):
Thanks for fixing this @giorgiofellipe

jordan-griffin (approved):

PR #171, opened 2024-08-27
Update README.md to use appropriate location fields

ajaysubra:
# Description

Setting lat and long on a profile doesn't impact the location of the profile on Klaviyo. Hence updated the example to use location properties that do impact the location if a developer was looking to set a user's location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,12 +205,12 @@ const profile: Profile = {
   title: 'CEO',
   organization: 'Muppets, Inc.',
   location: {
-    address1: '122 Main St',
+    address1: '666 Fake St.',
     address2: 'Apt 123',
-    city: 'Boston',
-    country: 'USA',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -205,8 +205,12 @@ const profile: Profile = {
...
   organization: 'Muppets, Inc.',
   location: {
-    latitude: 42.3601,
-    longitude: 71.0589,
+    address1: '122 Main St',
+    address2: 'Apt 123',
+    city: 'Boston',
+    country: 'USA',
+    region: 'MA',
+    zip: '02120',
```
kennyklaviyo, line 213: can we use a more fake address?

ajaysubra: Done.

dan-peluso (approved):

PR #173, opened 2024-09-12
Adopting modularization work

ajaysubra:
# Description

Pulling in the changes made on iOS for modularization. More comments in line.

# Check List

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -8,16 +8,6 @@ require Pod::Executable.execute_command('node', ['-p',
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
-# If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.
-# because `react-native-flipper` depends on (FlipperKit,...) that will be excluded
-#
-# To fix this you can also exclude `react-native-flipper` using a `react-native.config.js`
-# ```js
-# module.exports = {
-#   dependencies: {
-#     ...(process.env.NO_FLIPPER ? { 'react-native-flipper': { platforms: { ios: null } } } : {}),
-# ```
-flipper_config = ENV['NO_FLIPPER'] == "1" ? FlipperConfiguration.disabled : FlipperConfiguration.enabled
 
 linkage = ENV['USE_FRAMEWORKS']
 if linkage != nil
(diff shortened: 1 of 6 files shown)


example/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ajaysubra, line 1: Just to keep cocoapods happy since it kept complaining about this. We don't really need it since we don't publish this app to TF/public.

example/ios/Podfile
```diff
@@ -8,16 +8,6 @@ require Pod::Executable.execute_command('node', ['-p',
 platform :ios, min_ios_version_supported
 prepare_react_native_project!
 
-# If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.
```
ajaysubra, line 11: clean up.

example/ios/Podfile
```diff
@@ -32,15 +22,11 @@ target 'KlaviyoReactNativeSdkExample' do
 
   use_react_native!(
     :path => config[:reactNativePath],
-    # Enables Flipper.
```
ajaysubra, line 35: clean up. flipper is no longer supported

example/ios/Podfile
```diff
@@ -32,15 +22,11 @@ target 'KlaviyoReactNativeSdkExample' do
...
-    #
-    # Note that if you have use_frameworks! enabled, Flipper will not work and
-    # you should disable the next line.
-#    :flipper_configuration => flipper_config,
-    # An absolute path to your application root.
     :app_path => "#{Pod::Config.instance.installation_root}/.."
   )
 
+  pod 'KlaviyoSwift', :path => '../../../klaviyo-swift-sdk/'
```
ajaysubra, line 28: pointing to local version since the changes we made on iOS are not yet available via a release tag

ab1470: maybe add a comment in the code explaining the reason for this line, and instructing to remove it after we release SDK v4.0.0?

ab1470 (approved):

example/ios/Podfile
ajaysubra: Good callout. Done.

PR #175, opened 2024-09-19
CHNL-12052:  Update to include SDK name and version in network requests

ajaysubra:
# Description

⚠️ Merge after https://github.com/klaviyo/klaviyo-swift-sdk/pull/204 ⚠️ 

Does two things, 
1. Adds the plist `klaviyo-react-native-sdk-configuration.plist` to the iOS directory. 
2. Includes the above plist as a resource in the podspec.

Slightly unrelated but updated the breaking changes from the swift SDK and pushing them as well. FYI @ab1470 

## Test Plan

Tested locally by pulling in local version of Swift SDK and using the internal test app.

### DIFF
--- a/ios/KlaviyoBridge.swift
+++ b/ios/KlaviyoBridge.swift
@@ -31,10 +31,10 @@ public class KlaviyoBridge: NSObject {
   @objc
   public static var getEventTypesKeys: [String: String] {
       [
-        "VIEWED_PRODUCT": Event.EventName.ViewedProductMetric.value,
-        "STARTED_CHECKOUT": Event.EventName.StartedCheckoutMetric.value,
-        "OPENED_APP": Event.EventName.OpenedAppMetric.value,
-        "ADDED_TO_CART": Event.EventName.AddedToCartMetric.value
+        "VIEWED_PRODUCT": Event.EventName.viewedProductMetric.value,
+        "STARTED_CHECKOUT": Event.EventName.startedCheckoutMetric.value,
+        "OPENED_APP": Event.EventName.openedAppMetric.value,
+        "ADDED_TO_CART": Event.EventName.addedToCartMetric.value
       ]
   }
 
@@ -46,7 +46,6 @@ public class KlaviyoBridge: NSObject {
   @objc
   public static func initialize(_ apiKey: String) {
       KlaviyoSDK().initialize(with: apiKey)
-      KlaviyoCore.overrideSDKDefaults(name: "react_native", version: "9.9.9")
   }
 
   @objc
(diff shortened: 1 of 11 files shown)


ios/klaviyo-react-native-sdk-configuration.plist
```diff
@@ -0,0 +1,17 @@
...
+
+   Created by Ajay Subramanya on 9/18/24.
+   Copyright (c) 2024 Klaviyo. All rights reserved.
+-->
+<plist version="1.0">
+<dict>
+    <key>react_native_sdk_name</key>
+    <string>react_native</string>
+    <key>react_native_sdk_version</key>
+    <string>9.8.8</string>
```
evan-masseau, line 15: 9.8.8 ?

evan-masseau: I think we have a release doc pertaining to RN, lets make sure we update it to mention this is another spot the version needs to get incremented before release

klaviyo-react-native-sdk.podspec
```diff
@@ -13,6 +13,7 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => "13.0" }
   s.source       = { :git => "https://github.com/klaviyo/klaviyo-react-native-sdk.git", :tag => "#{s.version}" }
   s.source_files = "ios/**/*.{h,m,mm,swift}"
+  s.resources    = ["ios/klaviyo-react-native-sdk-configuration.plist"]
```
ndurell, line 16: There is no spm equivalent of this or that's not needed right?

ndurell (approved):

ios/klaviyo-react-native-sdk-configuration.plist
ajaysubra: 🤦🏾  I'm glad I used a crazy version for testing. Fixed it to 1.0.0 assuming it's going to be the next release. Good call, I update the confluence doc with this step.

klaviyo-react-native-sdk.podspec
ajaysubra: Hmm, didn't think about SPM. I think since we are including `KlaviyoSwift` SDK in the podspec and also since RN only supports cocoapods we should be good. Also, there is no package file in RN, there is no way to add this plist as a resource.

ios/KlaviyoBridge.swift
```diff
@@ -30,10 +31,10 @@ public class KlaviyoBridge: NSObject {
...
   public static var getEventTypesKeys: [String: String] {
       [
-        "VIEWED_PRODUCT": Event.EventName.ViewedProductMetric.value,
-        "STARTED_CHECKOUT": Event.EventName.StartedCheckoutMetric.value,
-        "OPENED_APP": Event.EventName.OpenedAppMetric.value,
-        "ADDED_TO_CART": Event.EventName.AddedToCartMetric.value
+        "VIEWED_PRODUCT": Event.EventName.viewedProductMetric.value,
+        "STARTED_CHECKOUT": Event.EventName.startedCheckoutMetric.value,
+        "OPENED_APP": Event.EventName.openedAppMetric.value,
+        "ADDED_TO_CART": Event.EventName.addedToCartMetric.value
```
ab1470, line 37: thanks! 🙏🏻

ab1470 (approved):

PR #177, opened 2024-09-23
CHNL-10057 Note react native firebase sdk capitalization issue

dan-peluso:
# Description

There's a change (issue?) in the react native firebase SDK we use where the push token from APNS is coming back as all caps, where the APNS API is not returning it in this format. This is just an added warning to the readme that describes the issue. 

# Check List

## Changelog / Code Overview

- updated readme
- 
## Related Issues/Tickets
CHNL-10057
https://github.com/invertase/react-native-firebase/issues/8022

### DIFF
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -54,26 +54,26 @@ there are two possible reasons for this:
 
 ### Deep links in push notifications not getting sent over to React Native layer when the iOS app is terminated
 
-When the iOS app is terminated (killed from the app switcher by swiping up), there is a bug in React Native that 
-prevents the native layer from calling the listeners set up in React Native to listen to incoming deep links. 
-The listener in question here is `Linking.getInitialURL()` which is expected to be called when a deep link is 
+When the iOS app is terminated (killed from the app switcher by swiping up), there is a bug in React Native that
+prevents the native layer from calling the listeners set up in React Native to listen to incoming deep links.
+The listener in question here is `Linking.getInitialURL()` which is expected to be called when a deep link is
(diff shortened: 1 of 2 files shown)


ab1470 (approved):

Troubleshooting.md
```diff
@@ -96,14 +96,19 @@ modified launch arguments, like so:
...
 ```
 
-This implementation is included in the example app in this repo and can be used for testing. 
+This implementation is included in the example app in this repo and can be used for testing.
 Make sure to use the URL scheme (rntest://) of the example app when testing.
 
+### iOS APNS Token Troubleshooting
+There is an open issue with [`@react-native-firebase/messaging`](https://github.com/invertase/react-native-firebase/issues/8022) where the SDK will uppercase any APNS token returned using `messaging().getAPNSToken()`.
+You can verify this by adding a log the `AppDelegate.m` file that prints the deviceToken (you will need to convert to a hex string).
+APNS Services is case-insensitive, so be sure you are not registering the same device token twice.
```
ajaysubra, line 113: Hmm, I'm not sure about this line. I've in the past not able to find official documentation to support what Apple thinks about push tokens. Also, there are some issues with getting a push synchronously after setting one so there may be some issues with asking developers to validate this. We handle this on our end so to keep integration light. My suggestion would be to not specify this but have this troubleshooting more of a FYI than anything actionable by the developer.

evan-masseau:
I am going to admin merge this. The CI issue is unrelated, and its just a markdown update that I would like to get into our docs.

PR #180, opened 2024-10-08
[CHNL-12052] Update to include SDK name and version for Android devices

dan-peluso:
# Description
By adding a manifest resource to the SDK, we can reflect in our Android Repo and check if these values exist, which tells our API the user is sending events through react-native and not android native SDK.

# Check List

## Changelog / Code Overview

- adding a new string resource folder for sdk name and version

## Test Plan

- tested with both the android and react example apps to see if the config was properly set, also updated a test to accommodate for the new error logging if the app context is not initialized

## Related Issues/Tickets
CHNL-12052

An important note about this is that technically a react-native SDK user could override this value and throw off our klaviyo config. Maybe we should add this to the README?

### DIFF
--- a/android/src/main/res/values/strings.xml
+++ b/android/src/main/res/values/strings.xml
@@ -1,0 +1,4 @@
+<resources>
+  <string name="klaviyo_react_native_sdk_version">1.0.0</string>
+  <string name="klaviyo_react_native_sdk_name">react_native</string>
+</resources>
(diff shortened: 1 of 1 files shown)


android/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<resources>
+  <string name="klaviyo_react_native_sdk_version">1.0.0</string>
```
evan-masseau, line 2: Nit: my suggestion would be to use a more generic name like 
`klaviyo_sdk_version_override`
`klaviyo_sdk_name_override`

ab1470 (approved):

PR #182, opened 2024-11-12
Renaming SDK Name and Version overrides to be more generic

dan-peluso:
If we're going to support different platforms overriding the native SDK values, we should probably make it named something less "react-native" specific. 

I'll open an ios PR to do this on that side as well

### DIFF
--- a/klaviyo-react-native-sdk.podspec
+++ b/klaviyo-react-native-sdk.podspec
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => "13.0" }
   s.source       = { :git => "https://github.com/klaviyo/klaviyo-react-native-sdk.git", :tag => "#{s.version}" }
   s.source_files = "ios/**/*.{h,m,mm,swift}"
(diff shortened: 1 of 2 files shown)


ab1470 (approved):

ajaysubra (approved):

klaviyo-react-native-sdk.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => "13.0" }
   s.source       = { :git => "https://github.com/klaviyo/klaviyo-react-native-sdk.git", :tag => "#{s.version}" }
   s.source_files = "ios/**/*.{h,m,mm,swift}"
-  s.resources    = ["ios/klaviyo-react-native-sdk-configuration.plist"]
```
ndurell, line 16: I think you have to rename this in the code too no?

ndurell: Oh nvm - that's in a different repo. 🤦🏻

PR #186, opened 2024-12-17
set editorconfig settings for Swift & objc files

ab1470:
# Description

This PR modifies the editorconfig to set the indent size to 4 for Swift & Objc file types, to reflect code style conventions

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,9 @@ root = true
 indent_style = space
 indent_size = 2
 
+[*.{swift,h,m,mm}]
+indent_size = 4
+
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true


ndurell (approved):

dan-peluso (approved):
Just curious what does this currently default to, and do we have to run anything for this to be applied to existing files

ab1470:
> Just curious what does this currently default to, and do we have to run anything for this to be applied to existing files

Prior to this change, the editorconfig file [set the indent size to 2](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/a9eec4ead63c61cfa426956750a548a70c1c8947/.editorconfig#L10) for all file types. This change sets it to 4 only for Swift & Objc files. 

Some of the existing Swift/Objc files have an indent width of 2. I didn't change them for this PR, but that can be addressed in a separate one. I just wanted to make this change now because I'm creating some new Swift/Objc files in my upcoming RN PR, and I wanted the new files to be formatted with 4 spaces

Issue #185, opened 2024-11-27
Compilation error for Android on latest React Native 0.76 with new architecture enabled


adrianso:
### Checklist

- [X] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/.github/CONTRIBUTING.md)
- [X] I have determined whether this bug is also reproducible in a vanilla project, such as the example app in this repository.
- [X] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [X] This issue hasn't been addressed in an [existing issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls)

### Description

```markdown
The SDK does not compile for Android in React Native 0.76.3 when new architecture is enabled.
```

### Expected behavior

```markdown
I expect the SDK to compile properly.
```

### Actual behavior

```markdown
I receive the following compilation error.

Failed to install the app. Command failed with exit code 1: ./gradlew app:installDebug -PreactNativeDevServerPort=8081

[...]

PR #194, opened 2025-01-30
Support for bridgeless NativeModule in New Architecture

evan-masseau:
# Description
This aims to add android support for the New Architecture by narrowing our current implementation to be exclusively a NativeModule that can run in bridgeless mode in new arch apps. I believe we had already stripped turbo module codegen from iOS in an earlier issue when we took out some podspec flags that were causing issues for customers. 

This will be a stepping stone toward proper TurboModule support. Since that migration may take longer, this can unblock folks who need new arch ASAP. 

# Check List

- [x] Are you changing anything with the public API? -- NO
- [ ] Are your changes backwards compatible with previous SDK Versions? -- So far yes, pending full regresssion testing

## Changelog / Code Overview

In our initial commit we selected a template intended to support Native _and_ Turbo Modules. Since we didn't fully implement TurboModules for version 1.0, the migration to New Architecture in RN 0.76.0 uncovered an issue for us: codegen was still running and erroring out on Android builds. To remedy this, I referenced the latest NativeModule project template to ensure we are up to date with best practices for making a _NativeModule_ available on the _New Architecture_ and upgraded the example app to 0.77.0 to catch up to the now-default architecture for new apps. Since we're still a Native Module, old arch support should be unaffected.

## Test Plan
Until I publish an alpha to npm, this version can be installed with:
`npm install git@github.com:klaviyo/klaviyo-react-native-sdk.git#ecm/bridgeless`

Test plans:
- [x]  Github CI covers example app compiling and building on each platform, and will now be defaulting to new arch.
- - [ ] It would be great if the CI could run against new and old arch, gotta look at the build scripts.
- [x]  Test against the bug repo provided in issue #185 for compilation
- [ ]  Manual tests of example app to confirm all SDK constants, setters, getters etc are bridging correctly (in progress)
- [ ]  Migrate internal test app to new architecture -- note this requires more package updates since we rely on more 3rd party libs for that app (in progress)

## Related Issues/Tickets
https://github.com/klaviyo/klaviyo-react-native-sdk/issues/185


.husky/pre-commit
```diff
@@ -1,9 +1,57 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
+#!/bin/sh
```
evan-masseau, line 1: This was from the new project template, it moves our git hooks to lefthook.yml. I don't have a strong opinion, but just figured I'd adopt it while we're updating

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 Kenny Tsui
+Copyright (c) 2025 Klaviyo
```
evan-masseau, line 3: Autogenerated by the template, no offense Kenny <3

android/build.gradle
```diff
@@ -79,20 +83,6 @@ android {
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
   }
-
```
evan-masseau, line 82: These gradle changes were necessary yet not sufficient for removing the turbomodule code gen from our project for now. When we do add TM support, they'll have to come back

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,8 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10-bin.zip
```
evan-masseau, line 3: Gradle update also appeared to be an important part of getting codegen to stop erroneously running, besides which we need it to support newer version of RN.

android/build.gradle
```diff
@@ -14,6 +14,11 @@ buildscript {
   }
 }
 
+def reactNativeArchitectures() {
```
evan-masseau, line 17: This just came from the template

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
```diff
@@ -14,135 +15,136 @@ import com.klaviyo.analytics.model.ProfileKey
 import java.io.Serializable
 import kotlin.reflect.KVisibility
 
-class KlaviyoReactNativeSdkModule internal constructor(private val context: ReactApplicationContext) :
```
evan-masseau, line 17: I flattened this class structure. We had this so that the base class could be different depending on old/new arch by using different source sets depending on architecture (also deleted the old/new source sets). We can assess and re-add what is needed when we add TM support

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkPackage.kt
```diff
@@ -1,39 +1,13 @@
...
-import com.facebook.react.TurboReactPackage
+import com.facebook.react.ReactPackage
 import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
-import com.facebook.react.module.model.ReactModuleInfo
-import com.facebook.react.module.model.ReactModuleInfoProvider
-import java.util.HashMap
+import com.facebook.react.uimanager.ViewManager
 
-class KlaviyoReactNativeSdkPackage : TurboReactPackage() {
```
evan-masseau, line 10: Stripped out references to TM for now, stick to the native module autoloader pattern referenced [here](https://reactnative.dev/docs/legacy/native-modules-android)

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainActivity.kt
```diff
@@ -51,25 +51,28 @@ class MainActivity : ReactActivity() {
           FirebaseMessaging.getInstance().token.addOnSuccessListener {
             Log.d("KlaviyoSampleApp", "Push token set: $it")
             Klaviyo.setPushToken(it)
-            Toast.makeText(
```
evan-masseau, line 54: Nothing substantively changed, must just be updated linter rules

example/android/build.gradle
```diff
@@ -1,14 +1,15 @@
...
-
         // Android Installation Step 1a: Set compatible minSdkVersion and compileSdkVersion
         minSdkVersion = 23 // Minimum supported by Klaviyo Android SDK
         targetSdkVersion = 33 // Typically minSdkVersion <= targetSdkVersion <= compileSdkVersion
         compileSdkVersion = 34 // Klaviyo Android SDK is compiled against 34
-
-        ndkVersion = "25.1.8937393"
-        kotlinVersion = "1.8.0"
+        buildToolsVersion = "35.0.0"
+        minSdkVersion = 24
```
evan-masseau, line 8: Our SDK supports 23, but RN 0.77 is 24+

example/android/gradle.properties
```diff
@@ -34,7 +32,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
+newArchEnabled=true
```
evan-masseau, line 35: 🔑

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
```diff
@@ -546,7 +546,7 @@
 					"$(inherited)",
 				);
 				INFOPLIST_FILE = KlaviyoReactNativeSdkExampleTests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 13.4;
```
evan-masseau, line 549: RN 0.77 bumped min ios support to 15.1. In some targets, I think we were missing any specifier, or had defaulted to higher than necessary

example/ios/KlaviyoReactNativeSdkExample/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,420 +1,354 @@
 {
-  "images": [
+  "images" : [
```
evan-masseau, line 2: 🤷 XCode wanted to reformat this file

example/react-native.config.js
```diff
@@ -1,10 +1,21 @@
...
+  project: {
+    ios: {
+      automaticPodsInstallation: true,
+    },
+  },
   dependencies: {
-    [pak.name]: {
+    [pkg.name]: {
       root: path.join(__dirname, '..'),
+      platforms: {
```
evan-masseau, line 13: This was a change I pulled from the new RN lib template, although it didn't seem to have any effect on our codegen issues.

lefthook.yml
```diff
@@ -1,14 +1,13 @@
...
-#      run: npx eslint {staged_files}
-#    types:
-#      glob: "*.{js,ts, jsx, tsx}"
-#      run: npx tsc --noEmit
-#commit-msg:
-#  parallel: true
-#  commands:
-#    commitlint:
-#      run: npx commitlint --edit
+pre-commit:
```
evan-masseau, line 1: Our pre-commit logic lives here now, still installed via husky tho

package.json
```diff
@@ -101,21 +109,13 @@
...
-    "node": ">= 18.0.0"
-  },
   "jest": {
     "preset": "react-native",
     "modulePathIgnorePatterns": [
       "<rootDir>/example/node_modules",
       "<rootDir>/lib/"
     ]
   },
-  "commitlint": {
```
evan-masseau, line 114: we weren't using this, and it enforces a rather annoying commit message format that I don't expect our devs to follow, for the most part because we typically squash merge and format commit messages at that time.

src/KlaviyoReactNativeSdk.ts
```diff
@@ -6,15 +6,8 @@ const LINKING_ERROR =
   '- You rebuilt the app after installing the package\n' +
   '- You are not using Expo Go\n';
 
-// @ts-expect-error
-const isTurboModuleEnabled = global.__turboModuleProxy != null;
-
-const KlaviyoReactNativeSdkModule = isTurboModuleEnabled
```
evan-masseau, line 12: This was supposed to be the basis of supporting NativeModules or TurboModules in the same lib. We'll probably have to revive this in the next phase so that we can maintain NM support when we add TM.

turbo.json
```diff
@@ -2,6 +2,7 @@
   "$schema": "https://turbo.build/schema.json",
   "pipeline": {
     "build:android": {
+      "env": ["ORG_GRADLE_PROJECT_newArchEnabled"],
```
evan-masseau, line 5: turbo build != turbo modules, but I did copy over this default env variable for new arch

src/Klaviyo.ts
```diff
@@ -1,5 +1,3 @@
-import { TurboModuleRegistry } from 'react-native';
```
evan-masseau, line 1: Renaming this file was probably the least obvious yet most important change. Evidently the `Native*` prefix is detected by codegen. 
I had expected that renaming this, in combination with the build.gradle changes above, _should have_ fixed our new arch code gen issues... but didn't. That's why I suspect there might have been an additional underlying bug in the RN/gradle libs but didn't have the time/bandwitdth to track it down myself. Besides which, this was a good time to update to latest RN version anyways so it wasn't really worth the effort.

LICENSE
kennyklaviyo: lolll

android/gradle.properties
```diff
@@ -12,10 +12,8 @@ org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryErro
 # org.gradle.parallel=true
 #Tue Dec 19 15:08:27 EST 2023
 KlaviyoReactNativeSdk_klaviyoAndroidSdkVersion=3.0.3
-KlaviyoReactNativeSdk_compileSdkVersion=31
-KlaviyoReactNativeSdk_kotlinVersion=1.8.0
+KlaviyoReactNativeSdk_kotlinVersion=1.7.0
```
kennyklaviyo, line 15: do we have to go down a kotlin version?

evan-masseau: Hm I didn't think much of this, but looking back, it seems we bumped that up from 1.7.0 in the past to 1.8.0 for RN versions <= 0.72.0 
Going to double check against #71 ...

.husky/pre-commit
```diff
@@ -1,9 +1,57 @@
...
 
-yarn lint-staged
+if [ "$LEFTHOOK_VERBOSE" = "1" -o "$LEFTHOOK_VERBOSE" = "true" ]; then
+  set -x
+fi
 
-if [ -f "package-lock.json" ];
-  then echo "package-lock.json is not allowed";
-  exit 1;
+if [ "$LEFTHOOK" = "0" ]; then
```
ajaysubra, line 7: > Fast. It is written in Go. Can run commands in parallel.
Okay!

android/gradle.properties
```diff
@@ -12,10 +12,8 @@ org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryErro
...
 KlaviyoReactNativeSdk_minSdkVersion=23
-KlaviyoReactNativeSdk_ndkversion=21.4.7075529
 KlaviyoReactNativeSdk_targetSdkVersion=31
-android.enableJetifier=true
-android.useAndroidX=true
```
kennyklaviyo, line 21: are these defaulted to true now? or incompatible with new arch

android/gradle.properties
evan-masseau: In that PR I wrote this (1.9.22 is the kotlin version of the android native SDK)

> 1.8 is compatible with 1.9.22 so we can stop here for now

Interesting because 1.7.0 is not causing any problems interfacing with our android SDK in my testing right now. I don't see any compelling reason to drop back to 1.7.0 though, so I will revert this line.

android/gradle.properties
evan-masseau: I was just noticing this as well. I'm not sure about jetifier, but we use androidX so that probably needs to come back

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainApplication.kt
```diff
@@ -10,19 +10,21 @@ import com.facebook.react.ReactPackage
 import com.facebook.react.defaults.DefaultNewArchitectureEntryPoint.load
 import com.facebook.react.defaults.DefaultReactHost.getDefaultReactHost
 import com.facebook.react.defaults.DefaultReactNativeHost
-import com.facebook.react.flipper.ReactNativeFlipper
+import com.facebook.react.soloader.OpenSourceMergedSoMapping
```
ajaysubra, line 13: I know they removed flipper support but what is soloader?

example/android/build.gradle
ajaysubra: so on native we support 23 but for RN it's only 24 onwards? NBD but checking.

example/android/gradle.properties
```diff
@@ -21,8 +21,6 @@ org.gradle.jvmargs=-Xmx2048m -XX:MaxMetaspaceSize=512m
 # Android operating system, and which are packaged with your app's APK
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
 android.useAndroidX=true
-# Automatically convert third-party libraries to use AndroidX
```
ajaysubra, line 24: any implication of removing this?

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
ajaysubra: I think same question as android, if someone where to update to these RN version they are letting go of lower iOS versions so it's not really us who is responsible for the higher min deployment target?

android/gradle/wrapper/gradle-wrapper.properties
kennyklaviyo: can this be `8.10.2` like the example android one uses??

evan-masseau: that would be nice, i'll check.

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
```diff
@@ -902,12 +901,8 @@
 					"-DFOLLY_USE_LIBCPP=1",
 					"-DFOLLY_CFG_NO_COROUTINES=1",
 				);
-				OTHER_LDFLAGS = (
-					"$(inherited)",
-					"-Wl",
-					"-ld_classic",
-				);
-				REACT_NATIVE_PATH = "${PODS_ROOT}/../../../node_modules/react-native";
+				OTHER_LDFLAGS = "$(inherited)";
```
ajaysubra, line 904: @belleklaviyo noticed this change in [her PR](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/188/files) I.E node module file path change (specifically `example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj`). Wondering if any local setup is going back and forth on this. Also, NBD since it's just the example app. May be versions of npm that we are using are different?

example/package.json
```diff
@@ -6,22 +6,24 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "build:android": "cd android && ./gradlew assembleDebug --no-daemon --console=plain -PreactNativeArchitectures=arm64-v8a",
-    "build:ios": "cd ios && xcodebuild -workspace KlaviyoReactNativeSdkExample.xcworkspace -scheme KlaviyoReactNativeSdkExample -configuration Debug -sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO"
+    "build:android": "react-native build-android --extra-params \"--no-daemon --console=plain -PreactNativeArchitectures=arm64-v8a\"",
```
ajaysubra, line 9: are these commands working now?

package.json
```diff
@@ -1,12 +1,22 @@
 {
   "name": "klaviyo-react-native-sdk",
-  "version": "1.1.1",
+  "version": "1.2.0-alpha.1",
```
ajaysubra, line 3: I'm sure you've thought about it but checkin in just to be sure - there is nothing breaking with these release correct?

src/Klaviyo.ts
ajaysubra: Neat find!

android/gradle.properties
evan-masseau: Looking back in git, it appears I added them pretty early on, but I'm not clear why. They weren't part of the original template, nor are they in the latest one. And I'm not sure why we'd need them, since the SDK's android files do not make use of androidx at all. 
The example app does have `useAndroidx=true` which makes sense because it has androidx dependency for the app and such. 
So I'm thinking this is safe to remove.

turbo.json
ajaysubra: what is turbo build? 😮‍💨

ajaysubra (approved):
LGTM some minor comments/questions.

turbo.json
evan-masseau: > Turborepo is a high-performance build system for monorepos, allowing you to manage multiple packages within a single repository efficiently. It optimizes build processes by caching and parallelizing tasks, resulting in faster development builds and improved performance.

I guess there's only so many adjectives in the english language lol. This was something that really boosted our CI performance especially iOS (but also caused those caching woes when we first added the GH actions)

android/gradle.properties
jordan-griffin, line 21: No support for RN libs on AndroidX?
Should we keep these in but set them explicitly to false?

example/android/build.gradle
evan-masseau: Well this is just the example app, since it was updated to 0.77, other RN packages probably won't work if I try to lower this to 23. I don't think it means the RN SDK will only support 24+, it'll just depend on the RN version of the host app.

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
evan-masseau: I want to confirm this, but I believe this is just dependent on the host app's version of RN. Since the example is now using 0.77, the min os version expectation has increased, but an RN app using 0.74 for example should still support 13.

package.json
evan-masseau: Shouldn't be, but version number is definitely TBD until we've finalized our regression testing

babel.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  presets: ['module:@react-native/babel-preset'],
+  presets: ['module:react-native-builder-bob/babel-preset'],
```
jordan-griffin, line 2: Is this necessary? Just confused on its relevance to this change

example/android/app/build.gradle
```diff
@@ -124,7 +127,6 @@ android {
 dependencies {
     // The version of react-native is set by the React Native Gradle Plugin
     implementation("com.facebook.react:react-android")
-    implementation("com.facebook.react:flipper-integration")
```
jordan-griffin, line 127: Guess there was some compatibility issue?

example/metro.config.js
```diff
@@ -1,45 +1,18 @@
...
 const root = path.resolve(__dirname, '..');
-const modules = Object.keys({ ...pak.peerDependencies });
 
 /**
  * Metro configuration
  * https://facebook.github.io/metro/docs/configuration
  *
  * @type {import('metro-config').MetroConfig}
  */
-const config = {
```
jordan-griffin, line 16: All this config stuff was for turbo modules or something?

package.json
```diff
@@ -140,6 +140,7 @@
       "prettier"
     ],
     "rules": {
+      "react/react-in-jsx-scope": "off",
```
jordan-griffin, line 143: wasthis? Does this allow us to declare JSX files without React imports?

babel.config.js
evan-masseau: There were quite a number of changes irrelevant to the turbomodules / new arch fix itself, that just came along with updating to the latest RN Native Module library template. I'm not certain about this one, but agree it'd be good to read up just to stay up to date with RN practices.

example/android/app/build.gradle
evan-masseau: I believe flipper support was fully removed from react native in 74 or 75. We weren't really using it anyway.

package.json
evan-masseau: its an eslint rule I doubt we'll ever be affected by, i believe it'll warn you off from trying to use react in jsx files if it isn't in scope... probably something you can get away with in theory because of the looser typing?

example/metro.config.js
evan-masseau: Yeah more templated stuff, this is a file we've never touched so I felt confident upgrading to the latest implementation.
I think build tools have been moving around over the releases since we initialized our repo, and so probably this stuff has been internalized into other rn packages.

android/gradle.properties
evan-masseau: [See above](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/194/files/ac39379d5a026c168733ed08eb68fc8bbb7ad5c2#r1937781180)

> Looking back in git, it appears I added them pretty early on, but I'm not clear why. They weren't part of the original template, nor are they in the latest one. And I'm not sure why we'd need them, since the SDK's android files do not make use of androidx at all.
The example app does have useAndroidx=true which makes sense because it has androidx dependency for the activities and notifications and such.
So I'm thinking this is safe to remove.

example/package.json
evan-masseau: I wasn't aware they weren't. They do work for me now yes.

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
evan-masseau: oh i didn't even notice the path difference, interesting. I meant to ask you about the "LDFLAGS" though

node version should be enforced by .nvmrc, so it _shouldn't_ be that. I'm not sure what applied the changes to this project file, it wasn't something I did explicitly (except the os version thing). But I wonder if the path difference is pointing to the SDK's `node_modules` dir vs the example app's. I'd think it should be the example app's, so the path I have here would be correct?

example/android/app/src/main/java/com/klaviyoreactnativesdkexample/MainApplication.kt
evan-masseau: https://github.com/facebook/SoLoader
Some kind of recursive library loader. Don't think it is related to flipper, just happens to be close in the alphabet for import order.

evan-masseau: which is funny because that explicitly says its needed for SDKs 23 and below, so IDK why they still have it in the template project if they bumped support to 24+.

android/gradle/wrapper/gradle-wrapper.properties
evan-masseau: ✅ and made them both use `bin` for the smaller package size too (`all` is packaged up with all the docs in case you're using gradle offline or something... which... how even?)

android/gradle.properties
evan-masseau: going back to 1.8.0

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
ajaysubra: image
Some linker settings! - https://developer.apple.com/documentation/xcode/build-settings-reference

ajaysubra: Yeah I think what you have now for path was what it was earlier. Don't worry much about it, we can figure this out if it happens to change again somehow.

jordan-griffin (approved):

android/gradle.properties
jordan-griffin: Ha, I should have refreshed before I commented! Thanks for clarifying

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkPackage.kt
dan-peluso: certainly looks cleaner

dan-peluso (approved):

PR #198, opened 2025-03-10
[CHNL-14728] In App Forms Support

dan-peluso:
# Description

- linking our new release to react native!
- also added to example app

# Check List

- [x] Are you changing anything with the public API?

## Changelog / Code Overview

[react-native-form.webm](https://github.com/user-attachments/assets/75e717e3-ef06-4f3c-afa1-69bf67c1120b)


example/src/KlaviyoReactWrapper.ts
```diff
@@ -19,7 +20,7 @@ export const initialize = async () => {
     // Alternate Android Installation Step 3
     // Alternate iOS Installation Step 3
     // Initialize the SDK with public key, if initializing from React Native
-    Klaviyo.initialize('YOUR_KLAVIYO_PUBLIC_API_KEY');
+    Klaviyo.initialize('UHZ3zG');
```
ajaysubra, line 23: nit picking - can you revert this?

kennyklaviyo, line 23: undo

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkPackage.kt
```diff
@@ -7,7 +7,10 @@ import com.facebook.react.uimanager.ViewManager
 
 class KlaviyoReactNativeSdkPackage : ReactPackage {
   override fun createNativeModules(reactContext: ReactApplicationContext): List<NativeModule> =
-    listOf(KlaviyoReactNativeSdkModule(reactContext))
+    listOf(
+      KlaviyoReactNativeSdkModule(reactContext),
+      KlaviyoFormsReactNativeSdkModule(reactContext),
```
ajaysubra, line 12: nice!

ios/KlaviyoFormsReactNativeSdk-Bridging-Header.h
```diff
@@ -0,0 +1,2 @@
+// KlaviyoFormsReactNativeSdk-Bridging-Header.h
```
ajaysubra, line 1: We had some issues with the bridging header in the main module. Wonder if that is now fixed with react native.

PR - https://github.com/klaviyo/klaviyo-react-native-sdk/pull/132/files
Issue - https://github.com/klaviyo/klaviyo-react-native-sdk/issues/129

ajaysubra (approved):
Neat!

example/src/KlaviyoReactWrapper.ts
dan-peluso: always forget something - sorry about that

ios/KlaviyoFormsReactNativeSdk-Bridging-Header.h
dan-peluso: Put this up for initial thoughts, planning to test on ios tomrrow

src/forms.tsx
```diff
@@ -0,0 +1,11 @@
+import { KlaviyoFormsReactNativeSdk } from './KlaviyoFormsReactNativeSdk';
+
+/**
+ * Klaviyo Forms
+ */
+
+export const KlaviyoForms = {
```
evan-masseau, line 7: Did you have a strong reason to go with a new const vs integrating with the existing `Klaviyo` interface, or was this just PoC so far? I don't have enough info yet to form an opinion on which approach we should go with.

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoFormsReactNativeSdkModule.kt
```diff
@@ -0,0 +1,36 @@
...
+    const val NAME = "KlaviyoFormsReactNativeSdk"
+  }
+
+  override fun getName(): String = NAME
+
+  override fun getConstants(): MutableMap<String, Any> = mutableMapOf()
+
+  @ReactMethod
+  fun registerForInAppForms() {
+    UiThreadUtil.runOnUiThread {
```
evan-masseau, line 26: 👍

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoFormsReactNativeSdkModule.kt
```diff
@@ -0,0 +1,36 @@
...
+      try {
+        Klaviyo.registerForInAppForms()
+      } catch (e: Exception) {
+        Log.d("KlaviyoReactNativeSdk", "Android unable to register for in app forms on main thread: ${e.message}")
```
evan-masseau, line 30: super nit: I think I'd want us to use the android SDK's implementation of Log, which would probably mean adding the core package as a gradle dependency, but thats fine.

belleklaviyo:
to get this to work for iOS we removed KlaviyoForms as subspec -- we can revisit if there is a better way to be able to install the KlaviyoForms module as a subspec later on but doing so rn to unblock other work

src/forms.tsx
dan-peluso: This was when I was following @ab1470 's spec, I think it was a better path for if we got the subpod thing working (or made it a new package to install) but not super clear

evan-masseau: Gotcha, yeah it's not clear to me from that spec how we'd conditionally exclude the forms specific code on android. Perhaps we need to spend more time investigating that

evan-masseau: Though similar to what Belle has found with subspec, maybe we punt on this modularization for MVP

src/Klaviyo.ts
```diff
@@ -18,4 +18,9 @@ export interface KlaviyoInterface
    * @param apiKey Your public API key
    */
   initialize(apiKey: string): void;
+
+  /**
+   * [ANDROID] used to track lifecycle interactions before the sdk is initialized
+   */
+  registerLifecycleCallbacks(): void;
```
evan-masseau, line 25: Don't they have to add this to their android application class in order for it to be called in time? If so, do we need to bridge the method to RN?

dan-peluso: To get forms to work yes, but I thought I'd add it in case there was other situations where a client wants the lifecycle monitoring behavior but maybe is unable to initialize the SDK immediately. Thinking out loud that's probably pretty edge case-y so I agree this can probably be removed for simplicity.

src/forms.tsx
evan-masseau: Coming back to this -- Given we're releasing 1.2.0 as a monolith package and not modularizing, I'd prefer to see `registerForInAppForms` merged into the `Klaviyo` interface, following the pattern of how we have split up interfaces already

src/KlaviyoFormsReactNativeSdk.ts
```diff
@@ -0,0 +1,20 @@
+import { NativeModules, Platform } from 'react-native';
```
evan-masseau, line 1: Do we still need to do this if not modularizing?

ios/KlaviyoFormsReactNativeSdk.mm
```diff
@@ -0,0 +1,25 @@
+// KlaviyoFormsReactNativeSdk.m
```
evan-masseau, line 1: Does this need to be its own file if we are not modularizing?

ios/KlaviyoFormsBridge.swift
```diff
@@ -0,0 +1,13 @@
+// KlaviyoFormsBridge.swift
```
evan-masseau, line 1: Do we still need to do this if not modularizing?

example/ios/Podfile.lock
```diff
@@ -1738,71 +1742,72 @@ SPEC CHECKSUMS:
...
+  ReactCodegen: 58a974a1a86362975fd49596480c5f0f17ee06a2
+  ReactCommon: e686c5766f0ebe5293be5a3957b833645cdac8ad
   SocketRocket: d4aabe649be1e368d1318fdf28a022d714d65748
-  Yoga: 93a4c16be50e6f29a7ef8e2b19505d7857c651fb
+  Yoga: c34725819ab0a5962e85455b9e56679b306910ee
 
 PODFILE CHECKSUM: fb4c6f946002a62b798ca073d975abf271898252
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.16.2
```
evan-masseau, line 1813: pod install should be run via gem exec, and pinned to v 1.15.2

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoFormsReactNativeSdkModule.kt
```diff
@@ -0,0 +1,32 @@
+package com.klaviyoreactnativesdk
```
evan-masseau, line 1: If not modularizing, do we need a new file for this?

evan-masseau (changes requested):
Leaving some change requests here, given we are not going with the modularizing approach, I'd like to keep things simpler. I care less about the internals, like whether we split up the native kt / objC / swift files, and more about the public API though.

android/build.gradle
```diff
@@ -119,6 +119,7 @@ dependencies {
   api "com.github.klaviyo.klaviyo-android-sdk:analytics:$klaviyoAndroidSdkVersion"
   api "com.github.klaviyo.klaviyo-android-sdk:push-fcm:$klaviyoAndroidSdkVersion"
   api "com.github.klaviyo.klaviyo-android-sdk:forms:$klaviyoAndroidSdkVersion"
+  api "com.github.klaviyo.klaviyo-android-sdk:core:$klaviyoAndroidSdkVersion"
```
evan-masseau, line 122: if we need core, we don't want to use `api`, we'd want to use `implementation`.

evan-masseau: but i also don't think we need add a `core` dependency

evan-masseau:
```suggestion
```

evan-masseau (approved):
Just took out the android core dependency via github, lgtm!
Ah and now that iOS CI is running, caught that the podfile.lock was missing the ios 4.2.0 version bump, adding that to keep testing this CI change...

PR #201, opened 2025-03-18
Bump React Native to 0.78.0

ab1470:
# Description

This PR bumps the React Native version to 0.78.0, to bring it in line with the test app (as of [this commit](https://github.com/klaviyo/klaviyo-react-native-test-app/commit/277fb3e4d0861f0defc002b604daf8a58e752b36))

# Check List

## Test Plan

I built/ran both the iOS and Android test apps and validated that they're working as expected.

## Related Issues/Tickets

These changes are required in order to do local testing of the upcoming RN release using the RN test app


belleklaviyo (approved):

evan-masseau (changes requested):
Take a look at this [upgrade helper tool](https://react-native-community.github.io/upgrade-helper/?from=0.77.0&to=0.78.0#RnDiffApp-package.json), there's a couple more 0.77 packages (likely quite minor) that we should also bump like
`babel-preset`
`metro-config`
`typescript-config`

They also recommend the command `npx @rnx-kit/align-deps --requirements react-native@[major.minor]`, last time I ran that it upgraded a couple things differently than the upgrade helper tool and I opted to go with the `align-deps` choices.

Thanks for doing this, it'll be good to keep test app and sdk on the same version

example/android/build.gradle
```diff
@@ -1,13 +1,10 @@
 buildscript {
     ext {
         // Android Installation Step 1a: Set compatible minSdkVersion and compileSdkVersion
-        minSdkVersion = 23 // Minimum supported by Klaviyo Android SDK
-        targetSdkVersion = 33 // Typically minSdkVersion <= targetSdkVersion <= compileSdkVersion
-        compileSdkVersion = 34 // Klaviyo Android SDK is compiled against 34
+        minSdkVersion = 24 // Minimum supported by Klaviyo Android SDK
```
evan-masseau, line 4: This comment can be removed, when i added it, RN supported something like 21+, but we had to narrow to 23+ bc our android SDK has min version of 23. Now its kinda the reverse since RN has bumped up their min supported android version.

```suggestion
        minSdkVersion = 24
```

example/android/build.gradle
```diff
@@ -1,13 +1,10 @@
...
+        targetSdkVersion = 35 // Typically minSdkVersion <= targetSdkVersion <= compileSdkVersion
+        compileSdkVersion = 35 // Klaviyo Android SDK is compiled against 35
```
evan-masseau, line 6: This comment can be removed, when i added it, RN supported something like 21+, but we had to narrow to 23+ bc our android SDK has min version of 23. Now its kinda the reverse since RN has bumped up their min supported android version.
And I think it is ok to just compile against latest. I should bump that up on Android SDK too. 
```suggestion
        minSdkVersion = 24
        targetSdkVersion = 35 // Typically minSdkVersion <= targetSdkVersion <= compileSdkVersion
        compileSdkVersion = 35
```

evan-masseau (approved):
Looks good!

PR #204, opened 2025-03-20
Script for testing SDK against past RN versions

evan-masseau:
It initializes an RN app from template, installs the Klaviyo SDK, and compiles it on ios/android.
Next steps could include
- make this script run against a number of RN versions in github actions
- add additional tests beyond the simple install+compile 

# Description

# Check List


evan-masseau:
outdated, and not really necessary, we can regenerate later if we want it

PR #206, opened 2025-03-25
updating codeowners

ajaysubra:
# Description
Updating codeowners

# Check List

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,6 +2,12 @@
 *   @klaviyo/mobile-app-growth
 
 # Push related files
-src/Push.ts                  @klaviyo/mobile-app-growth @klaviyo/push-team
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1,2 +1,7 @@
 # Default owner
-*   @klaviyo/mobile-push-team
+*   @klaviyo/mobile-app-growth
+
+# Push related files
```
ajaysubra, line 4: I kept the native module to default to mobile app growth since they are a single file and I think app growth should have ownership on them. I tried co-owners but it's easy to clearly define ownership so figured we'll start with this and see organically what files push ends up having to make changes to and update as we go.

njparadis (approved):

ab1470 (approved):

PR #216, opened 2025-06-04
CHNL-21453: Bridge unregister + config changes

belleklaviyo:
# Description
Bridges the changes to the forms API including the `configuration` to set the app session length and the `unregister` method

## Due Diligence

- [x] `Minor` Contains changes to the public API.

## Changelog / Code Overview
- Adds an optional `configuration` arg to the `registerForInAppForms` method to set the session length
- Adds `unregisterFromInAppForms` method

## Test Plan
I was unable to get the RN test app to target local swift SDK to test this (not yet released SDK version) so best I could do was test in the RN SDK example app.

Demo: Registers with config set to 10 second session, shows lifecycle events being fired to JS, shows app session length working as it creates a new one when the length of a session has elapsed, shows unregister method kills the webview

https://github.com/user-attachments/assets/3a87e149-484e-4329-8c3f-b9b56ab35ead

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-21453

### DIFF
--- a/android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
+++ b/android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
@@ -16,6 +16,7 @@ import com.klaviyo.analytics.model.ProfileKey
 import com.klaviyo.core.Registry
 import com.klaviyo.core.utils.AdvancedAPI
 import com.klaviyo.forms.registerForInAppForms
+import com.klaviyo.forms.unregisterFromInAppForms
 import java.io.Serializable
 import kotlin.reflect.KVisibility
 
@@ -68,6 +69,13 @@ class KlaviyoReactNativeSdkModule(
       } catch (e: Exception) {
         Registry.log.error("Android unable to register for in app forms on main thread", e)
       }
+    }
+  }
+
+  @ReactMethod
+  fun unregisterFromInAppForms() {
+    UiThreadUtil.runOnUiThread {
+      Klaviyo.unregisterFromInAppForms()
     }
   }
 
--- a/android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
+++ b/android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
@@ -60,10 +60,11 @@ class KlaviyoReactNativeSdkModule(
   }
 
   @ReactMethod
-  fun registerForInAppForms() {
+  fun registerForInAppForms(configuration: ReadableMap?) {
     UiThreadUtil.runOnUiThread {
       try {
-        Klaviyo.registerForInAppForms()
+        val sessionTimeoutDuration = configuration?.getDouble("sessionTimeoutDuration") ?: 3600.0 // Default to 60 minutes (3600 seconds)
+        Klaviyo.registerForInAppForms(com.klaviyo.forms.IAFConfiguration(sessionTimeoutDuration = sessionTimeoutDuration))
       } catch (e: Exception) {
         Registry.log.error("Android unable to register for in app forms on main thread", e)
       }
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
```diff
@@ -60,16 +61,28 @@ class KlaviyoReactNativeSdkModule(
   }
 
   @ReactMethod
-  fun registerForInAppForms() {
+  fun registerForInAppForms(configuration: ReadableMap?) {
```
ajaysubra, line 64: should the type be a mirror of the native sdk type?

ajaysubra: something like - https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/src/Event.ts#L77

ajaysubra: also, I see that `setProfile` doesn't take in a custom type and does `ReadableMap`. So I could be making an invalid suggestion.

ios/KlaviyoBridge.swift
```diff
@@ -51,8 +51,18 @@ public class KlaviyoBridge: NSObject {
...
     @MainActor
     @objc
-    public static func registerForInAppForms() {
-        KlaviyoSDK().registerForInAppForms()
+    public static func registerForInAppForms(configuration: [String: AnyObject]? = nil) {
+        guard let configurationLength = configuration?["sessionTimeoutDuration"] as? TimeInterval else {
+            KlaviyoSDK().registerForInAppForms()
+            return
+        }
+        KlaviyoSDK().registerForInAppForms(configuration: IAFConfiguration(sessionTimeoutDuration: configurationLength))
```
ab1470, line 59:
```suggestion
        if let configurationLength = configuration?["sessionTimeoutDuration"] as? TimeInterval {
            KlaviyoSDK().registerForInAppForms(configuration: IAFConfiguration(sessionTimeoutDuration: configurationLength))
        } else {
            KlaviyoSDK().registerForInAppForms()
        }
```
The end result here is the same, but for clarity I'd prefer an `if let` here instead of a `guard let`, for a similar reason as I describe [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/346#discussion_r2116512796). A `guard let` suggests that a condition is required in order for us to continue, and the early return is a fallback in case we don't meet that condition. In other words words, the "preferred" state is for the `guard let` to succeed, and the early return is just a safe way to handle the not-preferred scenario.

In this case, it's perfectly normal and expected to not have a `configurationLength` value, and we don't really have a (strong) preference here as to whether the `configurationLength` is provided or not, so I think `if let` is more appropriate here

ab1470 (approved):

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
ajaysubra: nvm this is the native module and not the api. I will show myself out!

evan-masseau: Here I was about to comment exactly the same thing

android/src/main/java/com/klaviyoreactnativesdk/KlaviyoReactNativeSdkModule.kt
```diff
@@ -60,16 +61,28 @@ class KlaviyoReactNativeSdkModule(
...
     UiThreadUtil.runOnUiThread {
       try {
-        Klaviyo.registerForInAppForms()
+        val sessionTimeoutDuration = configuration?.getDouble("sessionTimeoutDuration")
+        if (sessionTimeoutDuration != null) {
+          Klaviyo.registerForInAppForms(com.klaviyo.forms.IAFConfiguration(sessionTimeoutDuration = sessionTimeoutDuration))
```
evan-masseau, line 69: Late now, but couldn't we import the configuration data class so it doesn't have to be fully qualified here

PR #226, opened 2025-06-17
Fix configure local script

evan-masseau:
I was focused on the git branch/commit path, and didn't fix the local path. Consolidated to one method for updating the podfile for local, git branch, commit, or official version.

### DIFF
--- a/configure-sdk.sh
+++ b/configure-sdk.sh
@@ -64,12 +64,16 @@ function configure_podfile() {
 
   for dependency in "${dependencies[@]}"; do
     if [[ "$swift_sdk_version" =~ ^[0-9]+\.[0-9]+\.[0-9]+.*$ ]]; then
+      #Semantic version
       podfile_entry="pod '$dependency', '$swift_sdk_version'"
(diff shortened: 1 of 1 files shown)


configure-sdk.sh
```diff
@@ -42,6 +42,44 @@ function configure_local_properties() {
...
+      podfile_entry="pod '$dependency', :git => 'https://github.com/klaviyo/klaviyo-swift-sdk.git', :branch => '$swift_sdk_version'"
+    fi
+
+    nl=$'\n'
+    sed -i '' "${target_line}a\\${nl}  $podfile_entry${nl}" "$podfile"
+    echo "Added $dependency dependency to $podfile"
+    target_line=$((target_line + 1)) # Increment target line for next insertion
+  done
+}
+
```
evan-masseau, line 82: @ab1470 I was thinking I'd copy this over to the test app repo so we can use this to set version, branch, commit or local there too.

ab1470: yeah that sounds great

ab1470 (commented):
looks good, very handy

ab1470 (approved):

Issue #214, opened 2025-05-20
Event properties with number value 1/0 are sent as boolean true/false on iOS


justin-unterreiner-lulus:
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/.github/CONTRIBUTING.md)
- [x] I have determined whether this bug is also reproducible in a vanilla project, such as the example app in this repository.
- [x] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [x] This issue hasn't been addressed in an [existing issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls)

### Description

```markdown
When using the `Klaviyo.createEvent()` API, any property sent with a number value of `1` or `0` will be serialized and sent across the wire as boolean `true` or `false` value respectively. Other numbers do not have this issue. I believe this only affects iOS. I have confirmed that this does not occur in the vanilla iOS SDK (https://github.com/klaviyo/klaviyo-swift-sdk) and that it only occurs when using this React Native SDK.

This can be verified by using an HTTP proxy to inspect the JSON request payload sent to Klaviyo, as well as viewing the events via the Klaviyo profile page (e.g. https://www.klaviyo.com/profile/...).

This occurs for top level properties as well as properties on nested objects and arrays.
```

### Expected behavior

```markdown
Using `Klaviyo.createEvent()` with properties containing number values of `1` and `0` should be sent and shown in the Klaviyo profile screen as `1` and `0`.
```

### Actual behavior

```markdown
Using `Klaviyo.createEvent()` with properties containing number values of `1` and `0` are sent and shown in the Klaviyo profile screen as `true` and `false`.
```

### Steps to reproduce

```markdown
1. Use the following code snippet to trigger an event
2. Look at the profile in Klaviyo's profile page
3. Notice all the properties with value 1/0 are shown as true/false


    Klaviyo.createEvent({
      name: EventName.ADDED_TO_CART_METRIC,
      value: 42,
      properties: {
        quantity: 1,
        AddedItemQuantity: 1,
        number0: 0,
        number1: 1,
        number2: 2,
        number3: 3,
        nestedObject: {
          number0: 0,
          number1: 1,
          number2: 2,
          number3: 3,
        },
        nestedItemsArray: [
          {
            number0: 0,
            number1: 1,
[...]

PR #240, opened 2025-07-22
Prevent casting of 0/1 number properties to booleans

belleklaviyo:
# Description
In the RN to Swift transmission, the types can be wrapped in a NSNumber that then casts them into booleans if possible rather than preserving them as 0s and 1s. This extra helper  `fixNSNumberTypes` preserves that for any data structure we are passing as a property. Applied this fixer to both event properties and profile properties as that is where we are allowing any freeform structure that could be affected by this. Other properties like email/phone already have their own validation.

## Due Diligence

- [ ] I am confident these changes are implemented with feature parity across iOS and Android (if applicable).

## Changelog / Code Overview
- Adds fixNSNumberTypes to KlaviyoBridge to prevent 0/1 to bool casting

## Test Plan
Wanted to add unit tests but there is not much testing infrastructure in this repo at this point. Manually tested with two events (added to the example app) and saw on the Klaviyo Developer logs it did appear as the expected types:

image
image

## Related Issues/Tickets
Addresses https://github.com/klaviyo/klaviyo-react-native-sdk/issues/214
https://klaviyo.atlassian.net/browse/CHNL-21194

### DIFF
--- a/ios/KlaviyoBridge.swift
+++ b/ios/KlaviyoBridge.swift
@@ -94,7 +94,7 @@ public class KlaviyoBridge: NSObject {
             title: profileDict[ProfileProperty.title.rawValue] as? String,
             image: profileDict[ProfileProperty.image.rawValue] as? String,
             location: location,
-            properties: profileDict[ProfileProperty.properties.rawValue] as? [String: Any]
+            properties: fixNSNumberTypes(profileDict[ProfileProperty.properties.rawValue] as Any) as? [String: Any]
         )
 
         KlaviyoSDK().set(profile: profile)
@@ -174,7 +174,7 @@ public class KlaviyoBridge: NSObject {
 
         let event = Event(
             name: eventType,
-            properties: event["properties"] as? [String: Any],
+            properties: fixNSNumberTypes(event["properties"] as Any) as? [String: Any],
             value: event["value"] as? Double,
             uniqueId: event["uniqueId"] as? String
         )
@@ -228,6 +228,23 @@ public class KlaviyoBridge: NSObject {
         default:
             return Profile.ProfileKey.custom(customKey: str)
         }
+    }
+
+    // check to ensure bools and numbers are preserved as their types between the RN/Swift layers
+    static func fixNSNumberTypes(_ value: Any) -> Any {
+        if let num = value as? NSNumber {
+            if CFGetTypeID(num) == CFBooleanGetTypeID() {
+                return num.boolValue
+            }
+            return num.intValue
+        }
+        if let dict = value as? [String: Any] {
+            return dict.mapValues { fixNSNumberTypes($0) }
+        }
+        if let arr = value as? [Any] {
+            return arr.map { fixNSNumberTypes($0) }
+        }
+        return value
     }
 }
 
--- a/example/src/AppViewInterface.ts
+++ b/example/src/AppViewInterface.ts
@@ -15,6 +15,8 @@ import {
   getPushToken,
   registerForInAppForms,
   unregisterFromInAppForms,
+  sendTestEvent,
+  sendTestProfileWithBooleans,
 } from './KlaviyoReactWrapper';
 
 export interface AppViewInterface {
@@ -103,5 +105,15 @@ export const appViews: AppViewInterface[] = [
     title: 'Click to unregister from forms',
     color: '#d3ab10',
     onPress: unregisterFromInAppForms,
+  },
+  {
+    title: 'Click to send an event with booleans and 0/1s',
+    color: '#d3ab10',
+    onPress: sendTestEvent,
+  },
+  {
+    title: 'Click to send an profile property updatet with booleans and 0/1s',
+    color: '#d3ab10',
+    onPress: sendTestProfileWithBooleans,
   },
 ];
--- a/example/src/KlaviyoReactWrapper.ts
+++ b/example/src/KlaviyoReactWrapper.ts
@@ -5,6 +5,7 @@ import {
   type Profile,
   ProfileProperty,
   type FormConfiguration,
+  EventName,
 } from 'klaviyo-react-native-sdk';
 
 import {
@@ -202,6 +203,49 @@ export const sendRandomEvent = async () => {
       uniqueId: generateRandomName(5),
     };
     Klaviyo.createEvent(event);
+  } catch (e: any) {
+    console.log(e.message, e.code);
+  }
+};
+
+export const sendTestEvent = async () => {
+  try {
+    Klaviyo.createEvent({
+      name: EventName.STARTED_CHECKOUT_METRIC,
+      value: 42,
+      properties: {
+        true: true,
+        false: false,
+        number0: 0,
+        number1: 1,
+      },
+    });
+  } catch (e: any) {
+    console.log(e.message, e.code);
+  }
+};
+
+export const sendTestProfileWithBooleans = async () => {
+  try {
+    const myProfile: Profile = {
+      email: generateRandomEmails(),
+      phoneNumber: generateRandomPhoneNumber(),
+      externalId: generateRandomName(8),
+      firstName: 'Test',
+      lastName: generateRandomName(4),
+      organization: generateRandomName(5),
+      title: generateRandomName(6),
+      image: generateRandomName(5),
+      properties: {
+        // Test boolean properties in profile
+        isSubscribed: true,
+        hasConsented: false,
+        numberOfDogs: 0,
+        numberOfCats: 1,
+      },
+    };
+
+    Klaviyo.setProfile(myProfile);
   } catch (e: any) {
     console.log(e.message, e.code);
   }


dan-peluso (approved):

ios/KlaviyoBridge.swift
```diff
@@ -229,6 +229,23 @@ public class KlaviyoBridge: NSObject {
...
+            return num.intValue
+        }
+        if let dict = value as? [String: Any] {
+            return dict.mapValues { fixNSNumberTypes($0) }
+        }
+        if let arr = value as? [Any] {
+            return arr.map { fixNSNumberTypes($0) }
+        }
+        return value
+    }
```
ab1470, line 248: I had a feeling this code might be a little fragile, especially since we're dealing with `Any` types. I ran this through Gemini, and it suggested some changes. Here's the `fixNSNumberTypesNew` function it wrote:

```swift
func fixNSNumberTypesNew(_ value: Any?) -> Any? {
    guard let value = value, !(value is NSNull) else {
        return nil
    }

    switch value {
    case let num as NSNumber:
        // This check is to differentiate between a boolean and a number
        if CFGetTypeID(num) == CFBooleanGetTypeID() {
            return num.boolValue
        }
        // This is to differentiate between integer and floating point numbers
        let numberType = CFNumberGetType(num)
        switch numberType {
        case .float32Type, .float64Type, .floatType, .doubleType, .cgFloatType:
            return num.doubleValue
        default:
            return num.int64Value
        }
    case let dict as [String: Any?]:
        return dict.compactMapValues { fixNSNumberTypesNew($0) }
    case let arr as [Any?]:
        return arr.compactMap { fixNSNumberTypesNew($0) }
    default:
        return value
    }
}
```

Then I asked it to compare `fixNSNumberTypes` with `fixNSNumberTypesNew`, and here's what it said:

> Of course. Let's break down the two functions.
> 
> Both `fixNSNumberTypes` and `fixNSNumberTypesNew` are designed to solve a common problem when working with data that comes from Objective-C or JSON serialization in Swift. In these cases, all numbers (`Int`, `Double`, `Bool`) can be represented as `NSNumber` objects. These functions recursively traverse a data structure (of arrays and dictionaries) and convert generic `NSNumber` objects into specific Swift types like `Bool`, `Int64`, or `Double`.
> 
> ### `fixNSNumberTypes` (The older function)
> 
> This function has several significant issues:
> 
> 1.  **Incorrect Number Conversion**: It converts all non-boolean numbers to an `Int` using `num.intValue`. This is problematic because:
>     *   **Data Loss for Floats**: A floating-point number like `3.14` would be truncated to `3`.
>     *   **Potential Overflow for Large Integers**: `intValue` converts the number to a 32-bit integer. If the original number is larger than what a 32-bit integer can hold, it will lead to an incorrect value.
> 
> 2.  **No `nil` or `NSNull` Handling**: The function takes `Any` as an argument, so it cannot be called with `nil`. Furthermore, it doesn't check for `NSNull` values within collections. If an array or dictionary contained `NSNull`, it would be left as-is, which could cause crashes later.
> 
> ### `fixNSNumberTypesNew` (The newer function)
> 
> This function is a major improvement and is significantly more robust.
> 
> 1.  **Correct Number-Type-Detection**: Instead of converting all numbers to `Int`, it inspects the underlying type of the `NSNumber`:
>     *   It first checks if the `NSNumber` is a boolean, just like the old function.
>     *   If not, it uses `CFNumberGetType` to determine if the number is an integer or a floating-point value. It correctly converts floats/doubles to `Double` and integers to `Int64`. Using `Int64` is much safer as it supports the full range of 64-bit integers.
> 
> 2.  **Graceful `nil` and `NSNull` Handling**:
>     *   The function accepts an optional `Any?`, so it can handle `nil` input.
>     *   It has a `guard` statement at the top to check for both `nil` and `NSNull` values, returning `nil` for either.
>     *   When processing arrays and dictionaries, it uses `compactMap` and `compactMapValues`. This is a key improvement: these methods automatically filter out any `nil` values that result from the recursive calls, effectively cleaning the collections of `nil` and `NSNull` entries.
> 
> 3.  **Modern Swift Syntax**: It uses a `switch` statement to handle the different types, which is generally considered cleaner and more readable in Swift than multiple `if-let` statements.
> 
> ### Conclusion
> 
> **`fixNSNumberTypesNew` is vastly superior.** It is more robust, correct, and safe. It properly handles the different kinds of numbers, prevents data loss, and gracefully manages `nil` and `NSNull` values, which are common sources of crashes and bugs when processing external data. The older `fixNSNumberTypes` function is buggy and would be unsafe to use in a production environment.

Finally, I asked it to generate some unit tests, which I ran in an Xcode playground:

```swift
class FixNSNumberTypesTestsNew: XCTestCase {
    func testFixingBooleans() {
        XCTAssertEqual(fixNSNumberTypesNew(true as NSNumber) as? Bool, true, "Should correctly identify TRUE boolean")
        XCTAssertEqual(fixNSNumberTypesNew(false as NSNumber) as? Bool, false, "Should correctly identify FALSE boolean")
    }

    func testFixingIntegers() {
        XCTAssertEqual(fixNSNumberTypesNew(10 as NSNumber) as? Int64, 10)
        XCTAssertEqual(fixNSNumberTypesNew(0 as NSNumber) as? Int64, 0)
        XCTAssertEqual(fixNSNumberTypesNew(-5 as NSNumber) as? Int64, -5)
        let largeInt = Int64.max
        XCTAssertEqual(fixNSNumberTypesNew(largeInt as NSNumber) as? Int64, largeInt, "Should handle large integers")
    }

    func testFixingFloatingPointNumbers() {
        XCTAssertEqual(fixNSNumberTypesNew(3.14 as NSNumber) as? Double, 3.14)
        XCTAssertEqual(fixNSNumberTypesNew(-0.001 as NSNumber) as? Double, -0.001)
        XCTAssertEqual(fixNSNumberTypesNew(12345.6789 as NSNumber) as? Double, 12345.6789)
    }

    func testFixingStringsAndOtherTypes() {
        XCTAssertEqual(fixNSNumberTypesNew("hello") as? String, "hello", "Should leave strings unchanged")
        XCTAssertEqual(fixNSNumberTypesNew("") as? String, "", "Should handle empty strings")
    }

    func testHandlingNilAndNSNull() {
        XCTAssertNil(fixNSNumberTypesNew(nil), "Should correctly handle nil")
        XCTAssertNil(fixNSNumberTypesNew(NSNull()), "Should correctly handle NSNull")
    }

    func testFixingSimpleArray() {
        let inputArray: [Any] = [1 as NSNumber, true as NSNumber, 3.14 as NSNumber, "hello"]
        let result = fixNSNumberTypesNew(inputArray) as? [Any]
        XCTAssertNotNil(result)
        XCTAssertEqual(result?.count, 4)
        XCTAssertEqual(result?[0] as? Int64, 1)
       XCTAssertEqual(result?[1] as? Bool, true)
       XCTAssertEqual(result?[2] as? Double, 3.14)
       XCTAssertEqual(result?[3] as? String, "hello")
    }

   func testFixingArrayWithNilAndNSNull() {
       let inputArray: [Any?] = [1 as NSNumber, nil, true as NSNumber, NSNull(), "world"]
       let result = fixNSNumberTypesNew(inputArray) as? [Any]
       XCTAssertNotNil(result)
       XCTAssertEqual(result?.count, 3, "Should remove nil and NSNull values from array")
       XCTAssertEqual(result?[0] as? Int64, 1)
       XCTAssertEqual(result?[1] as? Bool, true)
       XCTAssertEqual(result?[2] as? String, "world")
   }

   func testFixingEmptyArray() {
       let inputArray: [Any] = []
       let result = fixNSNumberTypesNew(inputArray) as? [Any]
       XCTAssertNotNil(result)
       XCTAssertTrue(result?.isEmpty ?? false, "Should handle empty arrays")
   }

   func testFixingSimpleDictionary() {
       let inputDict: [String: Any] = [
           "integer": 10 as NSNumber,
           "boolean": false as NSNumber,
           "double": 99.99 as NSNumber,
           "string": "test"
       ]
       let result = fixNSNumberTypesNew(inputDict) as? [String: Any]
       XCTAssertNotNil(result)
       XCTAssertEqual(result?.count, 4)
       XCTAssertEqual(result?["integer"] as? Int64, 10)
       XCTAssertEqual(result?["boolean"] as? Bool, false)
       XCTAssertEqual(result?["double"] as? Double, 99.99)
       XCTAssertEqual(result?["string"] as? String, "test")
   }

   func testFixingDictionaryWithNilAndNSNull() {
       let inputDict: [String: Any?] = [
           "integer": 10 as NSNumber,
           "nothing": nil,
           "double": 99.99 as NSNumber,
           "nullValue": NSNull()
       ]
       let result = fixNSNumberTypesNew(inputDict) as? [String: Any]
       XCTAssertNotNil(result)
       XCTAssertEqual(result?.count, 2, "Should remove nil and NSNull key-value pairs from dictionary")
       XCTAssertEqual(result?["integer"] as? Int64, 10)
       XCTAssertEqual(result?["double"] as? Double, 99.99)
       XCTAssertNil(result?["nothing"])
       XCTAssertNil(result?["nullValue"])
   }

   func testFixingEmptyDictionary() {
       let inputDict: [String: Any] = [:]
       let result = fixNSNumberTypesNew(inputDict) as? [String: Any]
       XCTAssertNotNil(result)
       XCTAssertTrue(result?.isEmpty ?? false, "Should handle empty dictionaries")
   }

   func testFixingNestedStructure() {
       let input: [String: Any?] = [
           "level": 1,
           "nestedDict": [
               "level": 2 as NSNumber,
               "isNested": true as NSNumber,
               "value": 2.5 as NSNumber,
               "anotherNil": nil
           ],
           "nestedArray": [
               100 as NSNumber,
               false as NSNumber,
               NSNull(),
               [
                   "deeplyNested": true as NSNumber
               ]
           ]
       ]

       let result = fixNSNumberTypesNew(input) as? [String: Any]
       XCTAssertNotNil(result)

       // Test top level
       XCTAssertEqual(result?["level"] as? Int64, 1)

       // Test nested dictionary
       let nestedDict = result?["nestedDict"] as? [String: Any]
       XCTAssertNotNil(nestedDict)
       XCTAssertEqual(nestedDict?.count, 3)
       XCTAssertEqual(nestedDict?["level"] as? Int64, 2)
       XCTAssertEqual(nestedDict?["isNested"] as? Bool, true)
       XCTAssertEqual(nestedDict?["value"] as? Double, 2.5)
       XCTAssertNil(nestedDict?["anotherNil"])

       // Test nested array
       let nestedArray = result?["nestedArray"] as? [Any]
       XCTAssertNotNil(nestedArray)
       XCTAssertEqual(nestedArray?.count, 3, "Should remove NSNull from nested array")
       XCTAssertEqual(nestedArray?[0] as? Int64, 100)
       XCTAssertEqual(nestedArray?[1] as? Bool, false)

       let deeplyNestedDict = nestedArray?[2] as? [String: Any]
       XCTAssertNotNil(deeplyNestedDict)
       XCTAssertEqual(deeplyNestedDict?["deeplyNested"] as? Bool, true)
   }
}
```

All of the unit tests passed.

Note that this is Gemini's unrevised output, and I didn't thoroughly check the code or the tests, so take it with a grain of salt, but on first glance the explanation it gave makes sense and the code it wrote looks more robust

ab1470 (changes requested):

ios/KlaviyoBridge.swift
belleklaviyo: these were good unit tests to confirm with, but it was difficult to add them into the actual test file as there is not really an appropriate Tests dir or infrastructure besides a small test for the example app itself. I'd like to keep this PR contained as adding it in was getting bulky requiring bridging files and whatnot, so might explore adding tests in another PR. But as you said, running these unit tests themselves in an XCode playground seemed like good assurance, and I also retested it with actual requests and checked in the Developer logs.

ab1470: looks good to me. I spent a little time seeing if it were possible to add the unit tests as an XCPlayground in the RN project, but there's not really a way for the playground to access the KlaviyoBridge.swift file so I gave up. Thanks for making these changes.

ios/KlaviyoBridge.swift
```diff
@@ -229,6 +229,35 @@ public class KlaviyoBridge: NSObject {
...
+    // Ensure bools and numbers are preserved as their types between the RN/Swift layers
+    static func fixNSNumberTypes(_ value: Any?) -> Any? {
+        guard let value = value, !(value is NSNull) else {
+            return nil
+        }
+
+        switch value {
+        case let num as NSNumber:
+            // Differentiate between a boolean and a number
+            if CFGetTypeID(num) == CFBooleanGetTypeID() {
```
evan-masseau, line 242: Pretty impressive (also seems wild that we have to do this and react native doesn't handle that correctly already but 🤷)

Issue #243, opened 2025-08-05
Android crashes when object in properties has null value


krisidmisso:
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/.github/CONTRIBUTING.md)
- [x] I have determined whether this bug is also reproducible in a vanilla project, such as the example app in this repository.
- [x] If possible, I've reproduced the issue using the `master` branch or latest release of this package.
- [x] This issue hasn't been addressed in an [existing issue](https://github.com/klaviyo/klaviyo-react-native-sdk/issues) or [pull request](https://github.com/klaviyo/klaviyo-react-native-sdk/pulls)

### Description

```markdown
Using code below works on iOS on all cases, but in Android it crashes the app if one of the keys in the `properties` object is `null`:


Klaviyo.createEvent({
name: 'eventNameString',
properties: {
  key1: 'string',
  key2: 2,
  key3: null, // <-- this crashes the app with java exception
 }
});


Exception:

NullPointerException
null cannot be cast to non-null type java.io.Serializable
com.klaviyoreactnativesdk.KlaviyoReactNativeSdkModule in createEvent
```

### Expected behavior

```markdown
Should behave the same as in iOS. Null values should be handled by the SDK
```

### Actual behavior

```markdown
Crashes the app if object contains a key with null value
```

### Steps to reproduce

```markdown
1. Create a Klaviyo event
2. In the `properties` pass an object that has one of the key values `null`, similart to the example provided above
```

### The Klaviyo React Native SDK version information

2.0.1

### Environment Description

react-native: 0.79.3, android

evan-masseau:
Thank you for bringing this to our attention!
Looks like we're relying a little too much on the typescript types on these properties, which technically wouldn't allow null there, and not being type-safe enough in our kotlin bridging logic.

krisidmisso:
@evan-masseau thank you for the quick response.
Is there any reason why we should not send `null` properties in the `properties` object?

evan-masseau:
To be honest, I think it just hasn't come up before, would be curious what your use case is and I could put in a feature request for you. That'd be a little more involved since we'd have to update the native SDKs too. 

By the way I did confirm that on iOS we're already dropping those "invalid" null values with our typesafety checks. With this PR, Android would match that behavior.

krisidmisso:
[...]

PR #244, opened 2025-08-06
Safer type checking on event properties in the kotlin bridge.

evan-masseau:
# Description

Adding more type safety in the Kotlin bridge function for events. 

## Due Diligence

- [x] I have adjusted or added new test cases to team test docs, if applicable.

## Changelog / Code Overview

Improved type safety in Kotlin code responsible for bridging events from the React SDK to the Android SDK.

## Test Plan

Tested on android by using the sample app and creating events using a number of invalid types for metric, value, and properties (i.e. `null`, `undefined` or mis-matched types like using `"42"` instead of `42` for value, which is meant to be a `double`.

## Related Issues/Tickets

#243


dan-peluso (approved):
What you've added definitely makes sense, but I do wonder if we should audit any other calls while we're here. Maybe like adding an error log to the profile setting or validating email?

PR #258, opened 2025-09-02
Consume universal links branches of mobile SDKs

ab1470:
# Description
In order to get CI to run while we do dev work for Universal Links, we need to point the React Native SDK to consume unpublished versions of the iOS and Android SDKs. This is a temporary change that should be reverted before merging this feature branch into `master`.


evan-masseau:
> @evan-masseau any idea how to resolve [this](https://github.com/klaviyo/klaviyo-react-native-sdk/actions/runs/17408034993/job/49417673141?pr=258#step:9:83)?

@ab1470 let's start by doing a second build (just kicked one off) -- jitpack builds artifacts ondemand but almost always takes longer to build than the gradle's timeout, so the first attempt usually fails, then I have to retry a few minutes later. Lets see if that is it this time

evan-masseau (approved):

PR #284, opened 2025-11-12
Updated configure-sdk.sh to have a non-interactive mode

evan-masseau:
# Description
The prior implementation of `configure-sdk.sh` was interactive only. Its slick but not as convenient for shorthand. It also means AI agents can't run it, because their bash tools don't typically allow answering prompts. This will also make it easier to have the test app's configure script automatically configure the local SDK by invoke this script. 

It also now "remembers" the prior configured values for the native SDKs by keeping the prior version around in a comment, so for example you could do 
`./configure-sdk.sh -lia` to use local on bother platforms, and then
`./configure-sdk.sh -ia` will default you back to the version in the podspec / prior gradle.properties files.

## Due Diligence

- [ ] I am confident these changes are implemented with feature parity across iOS and Android (if applicable).

## Changelog / Code Overview
Added arguments for everthing so that you can just run it one-shot instead of having to be interactive. Makes it easier to run quickly, and it makes it possible for AI agents to execute the script for us. 

## Test Plan
  EXAMPLES:
```
    # Configure both platforms to default local paths
    ./configure-sdk.sh -l

    # Configure only Android to default local path (iOS skipped)
    ./configure-sdk.sh -la
    ./configure-sdk.sh -l -a

    # Configure only iOS to default local path (Android skipped)
    ./configure-sdk.sh -li
    ./configure-sdk.sh -l -i

    # Configure Android to custom path, skip iOS
    ./configure-sdk.sh -l -a ../../custom-android-sdk

    # Configure remote Android to master branch, skip iOS
    ./configure-sdk.sh -ra
    ./configure-sdk.sh -r -a

    # Configure remote Android version, skip iOS
    ./configure-sdk.sh -r -a 1.2.3

    # Configure both platforms to specific versions
    ./configure-sdk.sh -r --android=1.2.3 --ios=2.0.0

    # Configure to feature branch
    ./configure-sdk.sh -r --android=feat/new-feature --ios=skip

    # Skip pod install after configuration
    ./configure-sdk.sh -l --skip-pod-install
```

## Related Issues/Tickets
Part of ongoing effort to AI-ify the repos, and I want to enable the test-app repo to configure the SDK from its own `configure-sdk` script by calling to this one, when targeting local.

### DIFF
--- a/configure-sdk.sh
+++ b/configure-sdk.sh
@@ -275,7 +275,6 @@ function configure_android_gradle() {
       android_sdk_version="${android_sdk_version//\//~}-SNAPSHOT"
     fi
 
-    sed -i '' "s|^$property_key=.*|$property_key=$android_sdk_version|" "$property_file"
     echo "Targeting Android SDK version: $android_sdk_version"
   fi
 


configure-sdk.sh
```diff
@@ -87,23 +241,63 @@ function configure_podfile() {
...
-  echo
+  local sdk_path
+
+  # Use provided value or prompt
+  if [[ -n "$ANDROID_VALUE" ]]; then
+    # Replace "default" with actual default path
+    if [[ "$ANDROID_VALUE" == "default" ]]; then
+      sdk_path="$default_sdk_path"
+    else
+      sdk_path="$ANDROID_VALUE"
```
amber-klaviyo: is there any way to validate the input the user provides? or are we fine trusting whatever input we're given?

configure-sdk.sh
```diff
@@ -87,23 +241,63 @@ function configure_podfile() {
...
+    fi
+  else
+    read -rp "Enter the relative path to klaviyo-android-sdk (default: $default_sdk_path): " sdk_path
```
amber-klaviyo: when do we expect this path to be hit?

configure-sdk.sh
```diff
@@ -1,7 +1,161 @@
...
+    --ios=*)
+      IOS_VALUE="${1#*=}"
+      shift
+      ;;
+    --skip-pod-install)
+      SKIP_POD_INSTALL=true
+      shift
+      ;;
+    -*)
+      # Handle combined short flags (e.g., -la, -li, -lar)
```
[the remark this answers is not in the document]
evan-masseau: Correct, and intentional yeah. Tried to make it quite flexible. FWIW it's all quite quick and easy to rerun, so the stakes are low even if you do something unintentional

configure-sdk.sh
evan-masseau: No args, i e interactive mode

PR #294, opened 2025-12-10
Pull in published Android and iOS SDKs

belleklaviyo:
Just drafting this so we're ready


android/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,4 @@
 <resources>
-  <string name="klaviyo_sdk_version_override">2.2.0-alpha.1</string>
+  <string name="klaviyo_sdk_version_override">4.2.0-rc.1</string>
```
belleklaviyo: oops accidentally put the wrong version in [this](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/290)

ajaysubra (approved):

PR #300, opened 2026-01-09
Fix gradle issues

evan-masseau:
First fix for android issues


babel.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  presets: ['module:react-native-builder-bob/babel-preset'],
+  presets: ['module:@react-native/babel-preset'],
```
evan-masseau: And this should fix the flow syntax error, i think.

PR #333, opened 2026-04-06
[MAGE-444] feat(example/ios): differentiate silent from standard push with content-available

aahiltn:
# Description

`didReceiveRemoteNotification` fires for two distinct scenarios that should be handled differently:

- **Pure silent push** (`content-available: 1`, no `alert` in `aps`) — delivered silently with no banner. Forwarded to `PushNotificationsHelper.handleSilentPush` for background processing.
- **Standard push + content-available** — visible notification that also carries `content-available`. Handled by `willPresent`/`didReceive`; no silent-push forwarding needed.

Consistent with the differentiation implemented in the Flutter SDK example app (klaviyo/klaviyo-flutter-sdk#72).

## Due Diligence

## Changelog / Code Overview

- **`AppDelegate.mm`**: Add `didReceiveRemoteNotification:fetchCompletionHandler:` with differentiation logic. Pure silent pushes call `handleSilentPushWithUserInfo:` and (when `isDebug=YES`) show a UIAlertController. Standard pushes with `content-available` are ignored here and handled by `willPresent`/`didReceive`.
- **`PushNotificationsHelper.swift`**: Add `handleSilentPush(userInfo:)` as the designated hook for background processing on silent push delivery.

## Test Plan

1. Send a **pure silent push** (`content-available: 1`, no `alert`) → "Silent Push Received" alert appears in debug builds
2. Send a **standard push with content-available** (`content-available: 1` + `alert`) → notification banner shows normally, no silent-push alert
3. Send a **standard push without content-available** → normal notification behaviour unchanged

## Related Issues/Tickets

Resolves MAGE-444

### DIFF
--- a/example/ios/KlaviyoReactNativeSdkExample/PushNotificationsHelper.swift
+++ b/example/ios/KlaviyoReactNativeSdkExample/PushNotificationsHelper.swift
@@ -41,6 +41,12 @@ class PushNotificationsHelper: NSObject {
   }
 
   @objc
+  static func handleSilentPush(userInfo: NSDictionary) {
+    NSLog("[Klaviyo] Silent push received: %@", userInfo)
+    // Perform background work here, e.g. prefetch content or sync state.
+  }
+
+  @objc
   static func handleReceivingPush(
     response: UNNotificationResponse,
     completionHandler: @escaping () -> Void,
--- a/example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
+++ b/example/ios/KlaviyoReactNativeSdkExample/AppDelegate.mm
@@ -205,6 +205,10 @@ BOOL useNativeImplementation = YES;
                                                    animated:YES
                                                  completion:nil];
     }
+  } else if (apsPayload[@"content-available"]) {
+    if (isDebug) {
+      NSLog(@"Standard Push with Background Processing: %@", userInfo);
+    }
   }
 
   // You MUST call the completion handler within ~30 seconds.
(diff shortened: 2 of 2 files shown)


example/ios/KlaviyoReactNativeSdkExample/PushNotificationsHelper.swift
```diff
@@ -40,6 +40,12 @@ class PushNotificationsHelper: NSObject {
     KlaviyoSDK().set(pushToken: token)
   }
 
+  @objc
+  static func handleSilentPush(userInfo: NSDictionary) {
```
evan-masseau, line 44: just a nit: maybe move all the objective C code to swift code, its just so much more readable.

evan-masseau (approved):

example/ios/KlaviyoReactNativeSdkExample/PushNotificationsHelper.swift
aahiltn: Hmm, agree that it'd more readable. I'm happy to have this as a followup change at some point?

evan-masseau: Re-reading this, i feel like it'd be more useful if we showed RN developers how to use this data from react/typescript code. On the other hand, this whole app needs an overhaul so I suppose just leave this as is and we improve it later.

evan-masseau (approved):

PR #343, opened 2026-04-17
feat(example): rebuild JS layer with permission helpers, hooks, and app shell

evan-masseau:
# Description

Part 2 of 4 in the RN example-app overhaul chain for [MAGE-464](https://linear.app/klaviyo/issue/MAGE-464). Rebuilds the entire JS-layer of the example app:

- **`PermissionHelper`** — location + push permission flows wrapping `react-native-permissions` and `@react-native-firebase/messaging`
- **Four domain hooks** — `useAnalytics`, `useForms`, `useLocation`, `usePush`, each owning the state and SDK calls for one Klaviyo feature area
- **App shell (`App.tsx`)** — sectioned `SectionList` UI that consumes the hooks, wires UI buttons to handlers, loads the API key from env, and fails fast at module load if the key is missing
- **Cleanup** — removes the old `KlaviyoReactWrapper.ts` / `AppViewInterface.ts` / `RandomGenerators.ts` approach

## Due Diligence

## Changelog / Code Overview

| Area | Change |
|------|--------|
| `example/src/PermissionHelper.ts` (new) | Location + push permission request / status helpers; memoized Firebase availability probe |
| `example/src/hooks/useAnalytics.ts` (new) | Profile identity + attribute setters, structured Location support, events (test + Viewed Product) |
| `example/src/hooks/useForms.ts` (new) | In-app forms register/unregister |
| `example/src/hooks/useLocation.ts` (new) | Geofencing register/unregister, Get Current Geofences with modal display |
| `example/src/hooks/usePush.ts` (new) | Push permission + token flow; platform-aware `onTokenRefresh` (Android: FCM token; iOS: re-fetch APNs) |
| `example/src/App.tsx` (rewritten) | Sectioned UI wiring hooks to buttons; JS-first SDK init; fail-fast on missing `KLAVIYO_API_KEY` |
| `example/src/Styles.ts` | Expanded styles for the new sectioned layout |
| `example/src/env.d.ts` + `.env.example` | Env-var typing + example key placeholder |

## Key decisions

- **Fail fast on missing API key** — module-load throw with a clear setup message beats silently broken SDK calls or threading `isApiKeyConfigured` through every hook. Matches the Android example app pattern.
- **iOS location** — `WhenInUse → Always` upgrades are driven by separate user taps, never auto-chained, because iOS silently blocks the upgrade if requested in the same interaction as the initial grant.
- **iOS push** — `messaging().requestPermission()` triggers APNs registration; we use `getAPNSToken` (not `getToken`) because Klaviyo iOS expects raw APNs tokens.
- **Firebase availability probe** uses `messaging().app` — not just module resolvability — so we correctly detect when `[FIRApp configure]` was skipped due to missing plist.

## Test Plan

- [x] Hooks typecheck and lint clean
- [x] End-to-end behavior verified in the simulator on both iOS and Android

## Related Issues/Tickets

Part of [MAGE-464](https://linear.app/klaviyo/issue/MAGE-464)

**Chained PR series:**
1. theme + components (merged in #342)
2. **This PR** — permission helpers, hooks, app shell, JS-first init
3. native platform setup (iOS + Android Firebase push) — #345
4. docs — #346

🤖 Generated with [Claude Code](https://claude.com/claude-code)


example/src/PermissionHelper.ts
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  try {
+    let isAuthorized = false;
+
+    if (Platform.OS === 'android') {
+      const androidAuthStatus = await PermissionsAndroid.request(
+        PermissionsAndroid.PERMISSIONS.POST_NOTIFICATIONS
+      );
+      isAuthorized = androidAuthStatus === 'granted';
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 23b9733. Gated the `POST_NOTIFICATIONS` request on `Platform.Version >= 33` (Android 13+ / API 33 / Tiramisu). On older Android versions notifications are auto-authorized, so we short-circuit `isAuthorized = true` — matches the guard pattern already in `MainActivity.kt`.

example/src/PermissionHelper.ts
```diff
@@ -0,0 +1,374 @@
...
+          case RESULTS.UNAVAILABLE:
+            console.log('Location permission is not available on this device.');
+            return;
+          case RESULTS.BLOCKED:
+          case RESULTS.DENIED:
+            console.log('Location permission was blocked or denied.');
+            handleDeniedPermissionModal();
+            return;
+          case RESULTS.GRANTED:
+            return;
```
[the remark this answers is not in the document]
evan-masseau: Pushing back on this one — the `return` is intentional, not a bug. The example app drives location permission through two separate button taps on both iOS AND Android: tap 1 grants foreground (WhenInUse / FINE), which flips the state to `inUse` and reveals a second "Request Background Permission" button. Tap 2 then hits the background flow.

Auto-chaining on Android 11+ would immediately redirect the user to a Settings page for background access right after the foreground grant dialog — jarring UX for a demo. Mirroring iOS's mandatory two-tap pattern keeps the flow predictable cross-platform.

The JSDoc was misleading though — updated in 1460c85 to call out the two-tap design explicitly, and added an inline comment on the `case RESULTS.GRANTED: return` so it doesn't look accidental next time.

example/src/hooks/usePush.ts
```diff
@@ -0,0 +1,111 @@
...
+      (remoteMessage: { messageId?: string }) => {
+        console.log('[usePush] foreground message:', remoteMessage?.messageId);
+      }
+    );
+
+    return () => {
+      unsubscribeTokenRefresh();
+      unsubscribeMessage();
+    };
+  }, []);
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 3d2cbb8. Added `isApiKeyConfigured: boolean` parameter to `usePush` and gated the `useEffect` body with `if (!isApiKeyConfigured) return;` — mirrors the `useAnalytics` pattern. App.tsx updated to pass the flag through (rolled into the shell PR in the chain).

example/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,249 @@
...
+    const lon = longitude.trim() ? parseFloat(longitude) : undefined;
+    const location: Location = {
+      city: city || undefined,
+      country: country || undefined,
+      zip: zip || undefined,
+      latitude: Number.isFinite(lat) ? lat : undefined,
+      longitude: Number.isFinite(lon) ? lon : undefined,
+    };
+    Klaviyo.setProfile({ location });
+  };
```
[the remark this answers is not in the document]
evan-masseau: Fixed in 3d2cbb8. Added the same `hasLocation` guard to `handleSetLocation` as `handleSetProfile` has — early-returns when every field is blank so we don't ship an empty location object to the SDK.

example/src/PermissionHelper.ts
```diff
@@ -0,0 +1,380 @@
...
+        text: 'Cancel',
+        style: 'cancel',
+      },
+      {
+        text: 'Open Settings',
+        onPress: () => openAppSettings(),
+      },
+    ]
+  );
+};
```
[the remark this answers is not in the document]
evan-masseau: Pushing back — this duplication is transient across the chained PRs. `KlaviyoReactWrapper.ts` (along with `AppViewInterface.ts` and `RandomGenerators.ts`) gets deleted wholesale in the next PR in the chain (#344, the shell PR) when `App.tsx` is rebuilt on top of these hooks. There's no state where the merged chain has both files coexisting. Bugbot's seeing the partial view from this PR's diff against its base, not the composite view at the end of the chain.

example/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    const location: Location = {
+      city: city || undefined,
+      country: country || undefined,
+      zip: zip || undefined,
+      latitude: Number.isFinite(lat) ? lat : undefined,
+      longitude: Number.isFinite(lon) ? lon : undefined,
+    };
+    Klaviyo.setProfile({ location });
+  };
```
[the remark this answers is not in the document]
evan-masseau: Yeah this would be nice to extract for DRYness, even if it is small.

example/index.js
```diff
@@ -2,4 +2,18 @@ import { AppRegistry } from 'react-native';
...
+// no-op here; a real app would forward the payload to
+// Klaviyo.handleBackgroundMessage or similar.
+// The require is guarded so the app still runs when Firebase isn't linked.
+try {
+  const messaging = require('@react-native-firebase/messaging').default;
+  messaging().setBackgroundMessageHandler(async () => {});
+} catch {
+  // Firebase not available — no-op.
+}
+
```
evan-masseau: I still don't really understand this. Two questions
- What is this, would it confuse our app developers to have it, is the comment clear / accurate?
- Should this be in this PR, or the next one that is more push integration focused I thought

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
```diff
@@ -440,6 +441,19 @@
 			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
 			showEnvVarsInLog = 0;
 		};
+		70C59C6502E25FB8E9912FF8 /* [CP-User] [RNFB] Core Configuration */ = {
```
evan-masseau, line 444: whoa whats this, this normal?

example/src/hooks/useAnalytics.ts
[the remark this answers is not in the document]
evan-masseau: Fixed in de2d20a. Extracted into a `buildLocation()` helper that returns `Location | null`. Both handlers now delegate to it: `handleSetLocation` early-returns on null, `handleSetProfile` uses `?? undefined` when folding it into the aggregate profile.

example/index.js
evan-masseau: Good questions. Updated the comment in de2d20a to be much clearer.

**What it is:** @react-native-firebase/messaging requires you to register a background message handler at module-load time — before `AppRegistry.registerComponent`. If you don't, it warns on every cold start (`No task registered for key ReactNativeFirebaseMessagingHeadlessTask`).

**For Klaviyo push specifically, this is a no-op.** Klaviyo delivers notifications via APNs on iOS and via the FCM token registered with the Klaviyo backend on Android — neither path reaches an FCM background-data-message handler. This stub exists purely to silence the RNFB warning.

**Why this PR, not #345:** the call is JS-layer code and it's coupled to the `@react-native-firebase/messaging` dep that this PR adds to `package.json`. Moving it to #345 would split a single concern (wiring RNFB into the JS entry point) across two PRs. It's safe to keep here because the `try/catch` guards against Firebase not being linked.

example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
evan-masseau: Yep, this is normal and not something we authored. `pod install` auto-adds this `[CP-User] [RNFB] Core Configuration` build phase whenever the RNFirebase pods are installed — it's a bash script maintained by Invertase (the RNFB maintainers) that reads an optional `firebase.json` config file at build time and injects its settings into the Info.plist (analytics opt-outs, crashlytics flags, messaging auto-init, etc.).

We don't have a `firebase.json` in the example, so the script is effectively a no-op — it just runs, finds no config, and moves on. It got added to `project.pbxproj` as a side effect of the `pod install` we ran after adding the RNFB deps. It would regenerate on any future `pod install` anyway, so committing it keeps the Xcode project consistent between fresh checkouts and post-install state.

example/ios/Podfile.lock
```diff
@@ -4,12 +4,66 @@ PODS:
   - DoubleConversion (1.1.6)
   - fast_float (8.0.0)
   - FBLazyVector (0.81.5)
+  - Firebase/CoreOnly (12.10.0):
```
evan-masseau, line 7: **Firebase dep chunk starts here.** Everything from here through `PromisesObjC` (~line 60) is pulled in transitively by adding `@react-native-firebase/messaging` to `example/package.json`:

- `Firebase/CoreOnly`, `Firebase/Messaging`, `FirebaseCore`, `FirebaseCoreExtension`, `FirebaseCoreInternal`, `FirebaseInstallations`, `FirebaseMessaging` — the actual Firebase iOS SDK
- `GoogleDataTransport`, `GoogleUtilities/*` (7 subspecs) — Google's shared support libs that Firebase depends on
- `nanopb`, `PromisesObjC` — low-level deps (protobuf encoder, Obj-C promises) that GoogleDataTransport and FirebaseInstallations need

None of these are declared directly — they're all autolinked by CocoaPods when `pod install` sees `@react-native-firebase/*` in node_modules.

example/ios/Podfile.lock
```diff
@@ -2356,6 +2416,66 @@ PODS:
     - React-perflogger (= 0.81.5)
     - React-utils (= 0.81.5)
     - SocketRocket
+  - RNFBApp (24.0.0):
```
evan-masseau, line 2419: **RN bridging layer for Firebase.** `RNFBApp` and `RNFBMessaging` (line 2448) are the React Native wrapper pods that expose the Firebase iOS APIs to JS via @react-native-firebase. They sit on top of the native Firebase pods from the top of the file and are added automatically by `@react-native-firebase` autolinking.

These are the pods affected by the `$RNFirebaseAsStaticFramework = true` flag in `Podfile` — that flag tells RNFB to build these wrappers statically so they link cleanly against Firebase's static-only native cores when `use_frameworks!` is enabled.

example/ios/Podfile.lock
```diff
@@ -2356,6 +2416,66 @@ PODS:
...
+    - React-renderercss
+    - React-rendererdebug
+    - React-utils
+    - ReactCodegen
+    - ReactCommon/turbomodule/bridging
+    - ReactCommon/turbomodule/core
+    - RNFBApp
+    - SocketRocket
+    - Yoga
   - RNPermissions (5.4.4):
```
evan-masseau, line 2479: **react-native-permissions dep.** Just `RNPermissions` — no Permission-* subspecs appear because permission types are opt-in via the setup.rb invocation in `example/ios/Podfile`:

```ruby
node_require('react-native-permissions/scripts/setup.rb')
```

This project only enables the permission macros we actually use (LocationAlways, LocationWhenInUse, Notifications). Every other permission type stays disabled, keeping the binary size minimal.

example/src/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  // Build a Location from the current form fields, or null if every field is blank.
+  // Shared by handleSetLocation (location-only setProfile) and handleSetProfile
+  // (full-profile setProfile).
+  const buildLocation = (): Location | null => {
+    const lat = latitude.trim() ? parseFloat(latitude) : undefined;
+    const lon = longitude.trim() ? parseFloat(longitude) : undefined;
+    const hasLocation =
+      city || country || zip || Number.isFinite(lat) || Number.isFinite(lon);
```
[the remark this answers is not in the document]
evan-masseau: Fixed in fcb4542. `buildLocation` now trims each string field up front and uses the trimmed values for both the `hasLocation` truthy check and the emitted object. A whitespace-only `city` (etc.) now counts as blank and is dropped from the payload instead of being sent as `"   "`.

example/index.js
evan-masseau: claude didn't even sign their name

example/src/App.tsx
```diff
@@ -1,50 +1,377 @@
...
+      { title: 'Profile', data: ['profile'] },
+      { title: 'Events', data: ['events'] },
+      { title: 'In-App Forms', data: ['forms'] },
+      { title: 'Geofencing & Location', data: ['geofencing'] },
+      { title: 'Push Notifications', data: ['push'] },
+    ],
+    []
+  );
+
+  const renderSection = (sectionKey: string) => {
```
amber-klaviyo: renderSection is redefined on every render because it's a plain function inside App, and since all four domain hooks (useAnalytics, useForms, useLocation, usePush) live here, typing a single character in any text field causes the entire list to re-render. 

A possible fix is splitting each case into its own memoized component — e.g. `<ProfileSection analytics={analytics} />, <PushSection push={push} /> `— so only the section whose state actually changed re-renders. Could be worth improving here if we expect integrators to copy the pattern we use.

evan-masseau: Probably don't expected integrators to copy our view logic but yeah that still sounds annoying/bad.

example/src/PermissionHelper.ts
```diff
@@ -0,0 +1,391 @@
...
+ */
+export const checkPushPermissionStatus = async (): Promise<boolean> => {
+  const messaging = getMessagingInstance();
+  if (!messaging) {
+    return false;
+  }
+
+  try {
+    const status = await messaging.hasPermission();
+    const AuthorizationStatus = require('@react-native-firebase/messaging')
```
amber-klaviyo: The file already memoizes the Firebase require() call via _firebaseAvailable and getMessagingInstance() to avoid repeated module lookups (noted in the line 7 comment). But AuthorizationStatus is fetched with a separate inline require() inside both requestPushPermission and here in checkPushPermissionStatus, which contradicts that intent. Since AuthorizationStatus is a static property on the default export (not the instance), it could be captured once alongside the existing require() in getMessagingInstance — e.g. returning { instance: messaging(), AuthorizationStatus: messagingModule.AuthorizationStatus } — or just extracted to a module-scope lazy constant on first use.

amber-klaviyo (approved):
Two suggested improvements in the comments, but overall this looks solid, and seems like a major upgrade!

example/src/App.tsx
evan-masseau: Fixed in 265068f. Extracted each switch case into its own `React.memo`-wrapped component under `example/src/sections/` (`ProfileSection`, `EventsSection`, `FormsSection`, `GeofencingSection`, `PushSection`). Each takes only the hook result(s) it actually consumes, so a keystroke in one section no longer re-renders the others. `renderSection` in `App.tsx` is now a thin switch dispatcher.

example/src/PermissionHelper.ts
evan-masseau: Fixed in 265068f. Added a module-scope `getAuthorizationStatus()` helper that lazily captures `AuthorizationStatus` once (same pattern as `_firebaseAvailable` / `getMessagingInstance`) and guards with try/catch when Firebase isn't linked. Both `requestPushPermission` and `checkPushPermissionStatus` now call it instead of inlining `require()`.

example/src/App.tsx
evan-masseau: Amend updated (167b8d3) — also includes two additions bundled in the same commit: `Styles.ts` `toggleButtonText` base color changed from `secondaryText` to `primary` (inactive toggle now reads as clickable, not disabled), and `example/ios/Podfile.lock` regenerated via `pod install` with CocoaPods 1.16.2 (Firebase pods confirmed present throughout the chain).

example/src/PermissionHelper.ts
evan-masseau: Amend also updated to 167b8d3 — same commit includes the `Styles.ts` and `Podfile.lock` additions mentioned above. All four changes (section components, AuthorizationStatus hoist, toggle button color, regenerated Podfile.lock) land in the single amended 2-hooks HEAD. Cascade 4-native → 5-docs → overhaul all clean, no conflicts.

example/src/App.tsx
```diff
@@ -1,50 +1,128 @@
...
+import { useLocation } from './hooks/useLocation';
+import { usePush } from './hooks/usePush';
+
+// Local components / styles
+import { styles } from './Styles';
+import { SectionHeader } from './components/SectionHeader';
+
+// Section components — extracted so a keystroke in one section's field doesn't
+// re-render every other section. Each is wrapped in React.memo and only takes
+// the hook result(s) it needs as props.
```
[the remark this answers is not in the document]
evan-masseau: Good catch — you're right that the memo wrappers were inert with all hooks living in App. Fixed structurally by colocating each domain hook inside the section that consumes it: `usePush` in `PushSection`, `useForms` in `FormsSection`, `useLocation` in `GeofencingSection`, and `useAnalytics` in a merged `AnalyticsSection` (Profile fields + Events buttons share one hook instance, so an email typed in Profile is the same profile Events attribute to).

With that layout, `App` holds no hook-derived state — a state change inside one section's hook re-renders only that section, and sibling sections' hooks aren't re-invoked. Dropped all `React.memo` wrappers and any defensive `useCallback`/`useMemo` along with the props plumbing since they're no longer needed.

example/src/Styles.ts
```diff
@@ -189,7 +184,7 @@ export const styles = StyleSheet.create({
   },
   toggleButtonText: {
     ...typography.button,
-    color: colors.secondaryText,
+    color: colors.primary,
```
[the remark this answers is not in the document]
evan-masseau: Fixed in f57e012. Good catch — you're right that once `toggleButtonText` resolves to `colors.primary`, the `toggleButtonTextActive` override became a no-op. I deleted the rule entirely and removed both usages in `ToggleButtons.tsx`. The active/inactive distinction is still communicated by the container background (transparent vs `colors.activeButtonTint` via `toggleButtonActive`), so no visual change.

PR #352, opened 2026-04-27
feat(example): allow setting and persisting Company ID at runtime

evan-masseau:
# Description

Adds a debug-only settings affordance to the example app so the Klaviyo Company ID (public API key) used by the SDK can be swapped at runtime without rebuilding. Useful when bouncing between accounts/environments while exercising the SDK.

## Due Diligence

  - Example-app-only change; no SDK behavior change. No new unit tests added.
- [x] I have adjusted or added new test cases to team test docs, if applicable.
  - N/A — example-app dev affordance.
  - JS-only (with one Swift-toolchain config file) — both platforms get the feature for free.

  - Example-app-only change; nothing ships in the published SDK.

## Changelog / Code Overview

**New `useCompanyId` hook** (`example/src/hooks/useCompanyId.ts`)
- On mount, reads any persisted override from AsyncStorage. If one exists and differs from the build-time default, calls `Klaviyo.resetProfile()` then `Klaviyo.initialize(stored)` to rebind the SDK to that key.
- Exposes `companyId`, `isOverridden`, `changeCompanyId(newKey)`, `resetToDefault()`. Each mutator persists/clears the AsyncStorage entry and re-runs `resetProfile + initialize`.

**New `AppHeader`** (`example/src/components/AppHeader.tsx`)
- Fixed top bar with the app title and a `⚙` settings button.

**New `CompanyIdModal`** (`example/src/components/CompanyIdModal.tsx`)
- Displays the active Company ID, accepts a new value via TextInput (Save disabled when empty/unchanged), and conditionally renders a Reset to Default button when an override is in effect. Trims whitespace.

**App wiring** (`example/src/App.tsx`)
- Pulls in the hook, header, and modal. Module-scope `Klaviyo.initialize(KLAVIYO_API_KEY)` from `.env` is **intentionally unchanged** — the override is purely additive on top of the default integrator path that we document. Integrators reading the example shouldn't have to reason about the debug shim.

**Dependency**
- Adds `@react-native-async-storage/async-storage ^2.1.2` to the example app only.

**Unrelated cleanup: `.swiftlint.yml`**
- Small extra: adds a repo-root SwiftLint config that excludes `**/node_modules` and `**/Pods` (so we stop flagging vendored Swift code) and disables the arbitrary 250-line `type_body_length` rule. Drops lint findings from 6 (mostly inside dependencies) to 3 in our own bridge code. Pulled in here rather than its own PR because it's a few lines and was the lint surface I tripped over while wiring up this change.

## Test Plan

1. **Cold launch with default** — Fresh install / cleared storage. App boots; modal shows the `.env` Company ID; header chrome displays.
2. **Change ID via modal** — Tap ⚙, enter a different valid Company ID, Save. Modal closes, subsequent SDK calls (track event, identify) hit the new account. Verify in Klaviyo dashboard.
3. **Persistence across relaunch** — Force-quit the app and relaunch. The override persists; the modal shows the overridden ID with the `Default: …` hint underneath.
4. **Reset to Default** — Open the modal while overridden, tap Reset to Default. Modal closes, default Company ID is restored, and a subsequent relaunch confirms persistence cleared.
5. **Validation** — Save button is disabled when the input is empty or matches the active ID. Whitespace is trimmed.

## Related Issues/Tickets

None — debug affordance for the example app, intentionally untracked.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,7 @@
+excluded:
+  - '**/node_modules'
+  - '**/Pods'
+
+disabled_rules:
+  # Arbitrary 250-line class body limit; we'll judge complexity by other means.
+  - type_body_length
--- a/example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
+++ b/example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		ECAD002C2BC0AD0100000001 /* KlaviyoReactNativeSdkExampleTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECAD002D2BC0AD0100000002 /* KlaviyoReactNativeSdkExampleTests.swift */; };
 		13B07FBF1A68108700A75B9A /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 13B07FB51A68108700A75B9A /* Images.xcassets */; };
 		1BE92CB23FB415E631030426 /* Pods_KlaviyoReactNativeSdkExample.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EEC94C900FFF3B8C61C8D2B9 /* Pods_KlaviyoReactNativeSdkExample.framework */; };
 		346E4F992BC59CB200CBCFC1 /* NotificationService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 346E4F982BC59CB200CBCFC1 /* NotificationService.swift */; };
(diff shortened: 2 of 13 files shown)


evan-masseau:
bugbot run

belleklaviyo (approved):
ran into build issues while trying to run this (some vague pre-existing dependency giving a nonzero exit code) that I assume are local issues but rubber stamping as this otherwise looks ok and makes sense

evan-masseau:
> ran into build issues while trying to run this (some vague pre-existing dependency giving a nonzero exit code) that I assume are local issues but rubber stamping as this otherwise looks ok and makes sense

Could be the addition of the asyncstorage dependency, similar happened to me at first, had to just clear all the build caches etc that i could think of

PR #353, opened 2026-04-29
chore: bump to 2.4.0 (Android 4.4.0, Swift 5.3.1)

evan-masseau:
# Description

Bump the React Native SDK to 2.4.0, pinning native dependencies to Klaviyo Android 4.4.0 and Klaviyo Swift 5.3.1. Also fixes three latent bugs in the version-bump and SDK-configuration helper scripts that were silently no-op'ing or fighting each other when the podspec / Podfile / gradle.properties were left in non-default states.

## Due Diligence

- [x] `Minor` Contains changes to the public API.

## Changelog / Code Overview

**`f5cb130` — `fix(scripts): pin podspec / clean overrides on version bump`**

Two latent script bugs surfaced while preparing this release:

1. `bump-version.sh`'s `sed` for `KlaviyoSwift`/`Forms`/`Location` only matched an existing pinned form (`"Name", "x.y.z"`). It silently no-op'd against:
   - a bare line (`s.dependency "KlaviyoSwift"`), which is what lands when working against unreleased native versions, and
   - the `s.dependency "KlaviyoSwift" ##, "x.y.z"` form left behind by `configure-sdk.sh` after a local override.
2. The Android path left behind a stale `# saved:` comment after `bump-version.sh` ran. `configure-sdk.sh` later treated that comment as the source of truth on the next `--android gradle.properties`, silently reverting the bump.

Fix: `bump-version.sh` now shells out to `./configure-sdk.sh --android gradle.properties --ios podspec --skip-pod-install` first, scrubbing both stacks via the helper that owns the cleanup logic. Then it pins the podspec with a sed pair that handles bare / `##, "x"` / pinned uniformly, with a `grep` verify-or-`exit 1` guard so silent no-ops fail loudly going forward.

`configure-sdk.sh` also gains `KlaviyoSwiftExtension` override support so its podspec-restore path restores the default Extension pod entry, and its override path injects the matching `KlaviyoSwiftExtension` entry into the example app's Extension target. The Podfile gets a placeholder marker comment so the script knows where to insert.

**`5e41fbd` — `chore: bump native SDKs to Android 4.4.0 / Swift 5.3.1`**

Output of `./bump-version.sh -v 2.4.0 -a 4.4.0 -s 5.3.1`. All five Klaviyo pods (Core, Swift, Forms, Location, Extension) resolve to 5.3.1 in the example app's `Podfile.lock`.

Why 5.3.1 over 5.3.0: 5.3.1 patches a CocoaPods integration issue introduced in 5.3.0 — `KlaviyoSwiftExtension` drops its `KlaviyoCore` dependency and inlines shared types, fixing a link failure for apps using `KlaviyoSwiftExtension` in a Notification Service Extension target. ([5.3.1 release notes](https://github.com/klaviyo/klaviyo-swift-sdk/releases/tag/5.3.1))

**`76b8a2a` — `fix(scripts): use targeted pod update instead of pod install`**

`configure-sdk.sh` exists to swap the *source* of the Klaviyo pods (podspec ↔ branch ↔ commit ↔ local path). Plain `pod install` honors the `Podfile.lock` pin, so it errors out every time the source has changed:

```
[!] CocoaPods could not find compatible versions for pod "KlaviyoForms":
  In snapshot (Podfile.lock):
    KlaviyoForms (= 5.3.1)
  In Podfile:
    KlaviyoForms (from `…klaviyo-swift-sdk.git`, branch `<branch>`)
```

Switched both `configure-sdk.sh` and `bump-version.sh` to `pod update KlaviyoCore KlaviyoSwift KlaviyoForms KlaviyoLocation KlaviyoSwiftExtension` — re-resolves only the Klaviyo pods and leaves React Native / Firebase / etc. untouched. `bump-version.sh` previously updated only three of the five (omitting Core and Extension), letting Extension drift; now all five stay in lockstep.

## Test Plan

- [x] `./bump-version.sh -v 2.4.0 -a 4.4.0 -s 5.3.1` from a working tree with `configure-sdk.sh`-style overrides produces a clean podspec (pinned 5.3.1), clean `gradle.properties` (4.4.0, no `# saved:` artifact), clean `Podfile` (overrides removed, Extension placeholder restored), and a `Podfile.lock` with all five Klaviyo pods at 5.3.1.
- [x] `./configure-sdk.sh --ios <branch>` followed by `./configure-sdk.sh --ios podspec` round-trips cleanly — pod sources swap without `pod install` lockfile errors.
- [ ] Build example app on iOS simulator and confirm forms / push / analytics call paths.
- [ ] Build example app on Android emulator and confirm forms / push / analytics call paths.
- [ ] Verify SDK reports version `2.4.0` to backend in tracked events.

## Related Issues/Tickets

Part of MAGE-452.

### DIFF
--- a/configure-sdk.sh
+++ b/configure-sdk.sh
@@ -350,13 +350,21 @@ function configure_pods() {
   sed -i '' "/pod 'KlaviyoSwift'/d" "$podfile"
   sed -i '' "/pod 'KlaviyoForms'/d" "$podfile"
   sed -i '' "/pod 'KlaviyoLocation'/d" "$podfile"
+  sed -i '' "/pod 'KlaviyoSwiftExtension'/d" "$podfile"
 
   # Restore podspec
   sed -i '' 's/\(s\.dependency "KlaviyoSwift"\) ##, "\([^"]*\)"/\1, "\2"/' "$podspec"
   sed -i '' 's/\(s\.dependency "KlaviyoForms"\) ##, "\([^"]*\)"/\1, "\2"/' "$podspec"
   sed -i '' 's/\(s\.dependency "KlaviyoLocation"\) ##, "\([^"]*\)"/\1, "\2"/' "$podspec"
 
+  # Locate Extension target marker and prepare newline literal for sed insertions
+  local nl=$'\n'
+  local ext_marker_line
+  ext_marker_line=$(grep -n "# Insert override klaviyo-swift-sdk extension pod below this line when needed" "$podfile" | cut -d: -f1)
+
   if [[ "$swift_sdk_version" == "podspec" ]]; then
+    # Restore default extension pod entry (deleted above with the main pods)
+    sed -i '' "${ext_marker_line}a\\${nl}  pod 'KlaviyoSwiftExtension'${nl}" "$podfile"
     echo "Restored to $podspec"
     return
   fi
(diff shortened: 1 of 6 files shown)


configure-sdk.sh
```diff
@@ -404,13 +429,18 @@ fi
 if [[ "$ios_value" != "skip" ]]; then
   configure_pods "$ios_value"
 
-  # Handle pod install
+  # Refresh the Klaviyo pods. `pod update <names>` (vs plain `pod install`) is
+  # required because configure-sdk.sh is in the business of *changing* the pod
+  # source for those pods (podspec ↔ branch ↔ commit ↔ local path), and plain
+  # `pod install` refuses to bridge a Podfile.lock pin against a different
+  # source. Updating only the Klaviyo pods leaves the rest of Podfile.lock
+  # alone so unrelated deps (RN, Firebase, etc.) don't churn.
```
ab1470, line 437: nice addition!

ajaysubra (approved):

ab1470 (approved):

belleklaviyo (approved):

PR #397, opened 2026-08-04
feat(example): add Subscriptions section to the example app [MAGE-913]

Atharva8168:
# Description

Adds a **Subscriptions** section to the in-repo example app, demonstrating `Klaviyo.createSubscription` from MAGE-856.

> **Stacked PR.** Based on `asb/mage-856-bridge-subscription-method` so this diff shows only its own 5 files. GitHub will auto-retarget it to `feat/subscription-methods-sdk` when #396 merges and its branch is deleted.

## Due Diligence

  - Verified on a **physical device** by @Atharva8168: the Subscriptions section renders correctly, the warning is collapsed to its headline by default and expands/collapses on tap, and the spacing above the List ID field is right. Screenshots to follow separately.
  - The example app has **no test infrastructure** — no `test` script, no jest/RTL dependency, zero test files, and no `testID` convention anywhere in `example/src`. Adding either would mean inventing a convention and touching every shared component's prop interface, so this follows the existing sections in having none.
  - Pure React Native, no platform-conditional code.

  - Example-app only — no SDK surface change. The public API ships in #396.

## Changelog / Code Overview

Follows the established section pattern, so `App.tsx` keeps holding no feature state and each section owns its own hook.

- **`example/src/hooks/useSubscription.ts`** — list ID, custom source, per-channel consent, and the all-available-marketing flag. `buildChannels()` omits any channel with nothing selected rather than sending an empty consent array the native SDK would only warn about. Guards an empty list ID and an empty consent selection with `Alert.alert`.
- **`example/src/sections/SubscriptionSection.tsx`** — list ID + custom source fields, the broad-grant switch, and a `Collapsible` per-channel picker (email / SMS / WhatsApp). Leads with the set-an-identifier-first caveat via the existing `warningContainer` styles, since the native SDK drops a subscription whose channel has no matching identifier.
- **`example/src/components/ConsentToggle.tsx`** — a labelled `Switch`. `ToggleButtons` was deliberately **not** reused: it models a mutually exclusive either/or, whereas consent sub-types are each independently grantable.
- **`example/src/Styles.ts`** — three keys for the new toggle row; no per-file StyleSheet, matching the single-shared-`styles` convention.
- **`example/src/App.tsx`** — new `'subscriptions'` `SectionKey`, `SECTIONS` entry, and `renderSection` case.

Turning on all-available-marketing **disables** the per-channel picker rather than hiding it, so it's visible which choices are being superseded.

### Deviation from the ticket, flagged for your call

MAGE-913 specifies a listId field plus a Subscribe button hardcoded to email/marketing with "no channel/type UI". This builds the **full per-channel picker** instead, for parity with what the Flutter example app shipped in klaviyo/klaviyo-flutter-sdk#118. Straightforward to reduce to the literal scope if preferred.

Separately, the ticket's premise that scaffolding must be built from scratch is stale — it references `example/src/AppViewInterface.ts` and claims there is no `TextInput`/`useState`. That file no longer exists and the example app already provides `ProfileTextField`, `ToggleButtons`, `Collapsible`, and extensive `useState`. Noted on the ticket.

## Test Plan

Automated: `yarn typecheck`, `yarn lint`, `yarn test` all green (66 jest tests, unchanged by this PR).

Manual, on **both** platforms:

1. `yarn example android` / `yarn example ios` (requires `example/.env` with `KLAVIYO_API_KEY` — `App.tsx` throws at module scope without it).
2. Profile & Events → set an email.
3. Subscriptions → enter a real list ID, expand **Channel Consent**, enable Email — Marketing, tap **Subscribe**.
4. Confirm the profile appears on that list in-account.
5. Enter a custom source and re-subscribe; confirm `$source` reflects it.
6. Toggle **All available marketing** — confirm the per-channel switches grey out — and subscribe.
7. Negative: tap Subscribe with a list ID but no consent selected → "No consent selected" alert, nothing sent.
8. Negative: subscribe with no email/phone on the profile → native logs a warning and drops the request.

⚠️ On Android the SDK only transmits over an OS-**validated** network. An emulator behind a VPN/proxy/Private DNS or on a stale snapshot queues everything with "Send prevented while network unavailable". Use a physical device or a cold boot.

## Related Issues/Tickets

- Closes [MAGE-913](https://linear.app/klaviyo/issue/MAGE-913)
- Depends on [MAGE-856](https://linear.app/klaviyo/issue/MAGE-856) — #396, the bridge this consumes
- Flutter equivalent: [MAGE-912](https://linear.app/klaviyo/issue/MAGE-912) / klaviyo/klaviyo-flutter-sdk#118

## Summary by CodeRabbit

* **New Features**
  * Added a Subscriptions section to the app.
  * Added a subscription form for list IDs, optional sources, and marketing consent.
  * Added separate consent controls for email, SMS, and WhatsApp channels.
  * Added support for subscribing to all available marketing channels at once.
  * Added validation, accessible toggle labels, disabled states, and helpful profile warnings.
  * Added collapsible warning details with animated expand and collapse behavior.

### DIFF
--- a/example/src/Styles.ts
+++ b/example/src/Styles.ts
@@ -298,4 +298,27 @@ export const styles = StyleSheet.create({
   warningContainerWithBottomSpacing: {
     marginBottom: spacing.md,
   },
+  // Collapsible warning: tappable headline row + chevron. Mirrors the
+  // collapsible* styles so the affordance reads the same as the rest of the app,
+  // but keeps the warning palette.
+  warningHeaderRow: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+  },
+  warningHeaderText: {
+    ...typography.body,
+    color: colors.warningText,
+    fontWeight: '600',
+    // Shrink rather than push the chevron off the row on narrow screens.
+    flexShrink: 1,
+    paddingRight: spacing.sm,
+  },
+  warningChevron: {
+    ...typography.body,
+    color: colors.warningText,
+  },
+  warningCollapsibleContent: {
+    marginTop: spacing.xs,
+  },
 });
--- a/example/src/Styles.ts
+++ b/example/src/Styles.ts
@@ -291,4 +291,11 @@ export const styles = StyleSheet.create({
     marginLeft: spacing.sm,
     marginTop: spacing.xs / 2,
   },
+  // Opt-in spacing for a warning that sits above other content. Matches
+  // profileFieldContainer's bottom margin so the vertical rhythm lines up.
+  // Not folded into warningContainer, which renders as the only child in
+  // PushSection and would gain a trailing gap.
+  warningContainerWithBottomSpacing: {
+    marginBottom: spacing.md,
+  },
 });
--- a/example/src/Styles.ts
+++ b/example/src/Styles.ts
@@ -253,6 +253,23 @@ export const styles = StyleSheet.create({
     fontWeight: '500',
   },
 
+  // Consent toggle styles
+  consentToggleRow: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    paddingVertical: spacing.xs,
+  },
+  consentToggleLabel: {
+    ...typography.body,
+    color: colors.text,
+    flexShrink: 1,
+    paddingRight: spacing.sm,
+  },
+  consentToggleLabelDisabled: {
+    color: colors.secondaryText,
+  },
+
   // Warning message styles
   warningContainer: {
     backgroundColor: colors.warningBackground,
--- a/example/src/sections/SubscriptionSection.tsx
+++ b/example/src/sections/SubscriptionSection.tsx
@@ -12,7 +12,12 @@ export function SubscriptionSection() {
 
   return (
     <View style={styles.section}>
-      <View style={styles.warningContainer}>
+      <View
+        style={[
+          styles.warningContainer,
+          styles.warningContainerWithBottomSpacing,
+        ]}
+      >
         <Text style={styles.warningText}>
           Set an email or phone number before subscribing.
         </Text>
--- a/example/src/sections/SubscriptionSection.tsx
+++ b/example/src/sections/SubscriptionSection.tsx
@@ -1,0 +1,92 @@
+import { Text, View } from 'react-native';
+
+import { useSubscription } from '../hooks/useSubscription';
+import { styles } from '../Styles';
+import { ActionButton } from '../components/ActionButton';
+import { Collapsible } from '../components/Collapsible';
+import { ConsentToggle } from '../components/ConsentToggle';
+import { ProfileTextField } from '../components/ProfileTextField';
+
+export function SubscriptionSection() {
+  const subscription = useSubscription();
+
+  return (
+    <View style={styles.section}>
+      <View style={styles.warningContainer}>
+        <Text style={styles.warningText}>
+          Set an email or phone number before subscribing.
+        </Text>
+        <Text style={styles.warningSubtext}>
+          The native SDK drops a subscription whose channel has no matching
+          identifier on the profile. Use Profile &amp; Events above to set one.
+        </Text>
+      </View>
+      <ProfileTextField
+        label="List ID"
+        value={subscription.listId}
+        onChangeText={subscription.setListId}
+        placeholder="ABC123"
+      />
+      <ProfileTextField
+        label="Custom Source (optional)"
+        value={subscription.customSource}
+        onChangeText={subscription.setCustomSource}
+        placeholder="Checkout screen"
+      />
+      <ConsentToggle
+        label="All available marketing"
+        value={subscription.allMarketing}
+        onValueChange={subscription.setAllMarketing}
+      />
+      {/*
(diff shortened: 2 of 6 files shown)


aahiltn:
mind grabbing some screenshots/videos for this?

example/src/sections/SubscriptionSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <View style={styles.section}>
+      <View style={styles.warningContainer}>
+        <Text style={styles.warningText}>
+          Set an email or phone number before subscribing.
+        </Text>
+        <Text style={styles.warningSubtext}>
+          The native SDK drops a subscription whose channel has no matching
+          identifier on the profile. Use Profile &amp; Events above to set one.
+        </Text>
+      </View>
```
belleklaviyo: Image

tiniest nit in the world, feel free to ignore but could we add a tiny space between list ID and the yellow alert above lol

belleklaviyo (approved):
built and run on iOS sim and tested some basic list subscription flows, lgtm!

example/src/sections/SubscriptionSection.tsx
Atharva8168: Not a nit at all — it was flush against the field, thanks for the screenshot. Fixed in `8ed667f`.

Added an opt-in `warningContainerWithBottomSpacing` modifier using the same `spacing.md` as `profileFieldContainer`, so the gap matches the rhythm of the fields below rather than being an arbitrary value.

Kept it as a modifier instead of just adding `marginBottom` to `warningContainer` itself, because `PushSection` renders that container as the only child of its branch and would have picked up a stray trailing gap.

aahiltn:
super super small nit, and feel free to push back:
i feel like the email / phone number warning should be a bit smaller (either with no description) or collapsable in some way. Mostly because I don't see extensive testing happening with this example app, and id rather keep this new section uniform with the rest. 

Nonblocking though. 

image

aahiltn (approved):
Tested iOS and Android Pathes on physical devices. small ui nit, but looks good.

Image

Image

Atharva8168:
Not pushing back at all — you were right, and thank you for taking the time to screenshot it. Between this and Belle's catch on the spacing, the review attention on this section has been genuinely useful; both were things I would not have seen from the diff alone.

Fixed in `57162e6`. The warning is now **collapsed by default** to a single headline line with a small chevron, and tapping it expands the explanation:

```
⚠️ Set an email or phone number first        ▸
```

```
⚠️ Set an email or phone number first        ▾
   The native SDK drops a subscription whose channel has no
   matching identifier on the profile. Use Profile & Events
   above to set one.
```

A new `CollapsibleWarning` component reuses `Collapsible`'s chevron glyphs (`▸`/`▾`) and its `easeInEaseOut` animation, so the affordance reads exactly like the collapsibles elsewhere in the app — which is the uniformity you were after. It keeps the warning palette rather than the neutral field-group styling, since it is still a caveat rather than a field group. I also shortened the headline now that the detail is one tap away.

It carries its own bottom spacing so Belle's fix for the gap above the List ID field is preserved.

Two things worth flagging honestly:

- **I have not seen this rendered.** tsc, eslint, and the 75 tests pass, but my emulator is down and port 8081 is occupied, so it is verified structurally rather than visually. Given you both found the previous issues *from screenshots*, that is a real gap.

example/src/Styles.ts
```diff
@@ -274,4 +291,34 @@ export const styles = StyleSheet.create({
...
+    marginBottom: spacing.md,
+  },
+  // Collapsible warning: tappable headline row + chevron. Mirrors the
+  // collapsible* styles so the affordance reads the same as the rest of the app,
+  // but keeps the warning palette.
+  warningHeaderRow: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+  },
```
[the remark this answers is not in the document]
Atharva8168: Valid catch — the row was the only press target and sat at roughly one text line, well under the 44pt the rest of the app uses (`profileFieldButton`, `profileFieldInput`). Fixed in `ba567b8`.

I went with `hitSlop` rather than the suggested `minHeight: 44`, though:

```tsx
hitSlop={{ top: 12, bottom: 12, left: 0, right: 0 }}
```

`minHeight` would have grown the collapsed warning by roughly 40%, which works directly against why it was collapsed in the first place — [@aahiltn asked for it to be smaller](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/397#issuecomment-5244193869) so the section stays uniform with the rest of the example app. `hitSlop` extends the target into the container's existing `spacing.md` padding, which is dead space either way, so the touch target clears 44pt while the visual stays a single line. Both concerns satisfied instead of trading one for the other.

Atharva8168:
Update on the verification gap I flagged above: I have now run this on a physical device and confirms the collapsible warning behaves as intended — collapsed to the single headline by default, chevron expands and collapses the detail, and the spacing above the List ID field is correct.

So both review points are confirmed working on-device, not just structurally:
- the warning is compact and uniform with the rest of the section ([@aahiltn](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/397#issuecomment-5244193869))
- the expand/collapse tap target is comfortable in practice, which was the concern behind the `hitSlop` change

PR #394, opened 2026-08-03
chore(deps): upgrade React Native 0.81.5 -> 0.86.2 (MAGE-919)

glenn-klaviyo:
# Description

Moves the SDK off end-of-life React Native 0.81.5 onto 0.86.2, the latest stable release. RN 0.81 sits outside React Native's active-support window (0.84–0.86), so it no longer receives upstream security or bug fixes.

## Due Diligence

- [x] I have tested this on a simulator/emulator or a physical device, on iOS and Android (if applicable).
- [ ] I have added sufficient unit/integration tests of my changes.
- [ ] I have adjusted or added new test cases to team test docs, if applicable.
- [x] I am confident these changes are implemented with feature parity across iOS and Android (if applicable).

Notes:
- No new tests: this is a dependency/toolchain upgrade with **zero SDK source changes**. The existing 57 unit tests all pass unchanged, including `src/__tests__/index.test.tsx`, which hand-mocks the `react-native` module and therefore would have broken had the bridge surface shifted.
- Verified on the iOS Simulator (iPhone 17, iOS 26.5) and an Android emulator (API 37): both build, launch, render, and initialize the native SDK.
- **Push verified on physical devices**: push token registration and push notification delivery both confirmed. Neither is obtainable on a simulator/emulator (iOS Simulator cannot get a genuine APNs token, and FCM needs Play Services), so this was validated on real hardware.

## Release/Versioning Considerations

- [x] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [ ] `Minor` Contains changes to the public API.
- [ ] `Major` Contains **breaking** changes.
- [x] Contains readme or migration guide changes.
- [x] This is planned work for an upcoming release.

The public API is unchanged. `peerDependencies` moves from `"*"` to a real range, but the floor (`>=0.68.0`) is set **below** the oldest version the README has ever documented as working (0.68.7), so no existing consumer's install breaks. Everything else is dev/CI tooling or example-app-only.

README change is a single line recording the tested RN version.

### JS dependencies

| | From | To |
|---|---|---|
| `react-native` | 0.81.5 | **0.86.2** |
| `react` | 19.1.0 | **19.2.8** (RN 0.86.2 peers `^19.2.3`) |
| `@react-native/{babel-preset,eslint-config,new-app-screen,metro-config,typescript-config}` | 0.81.5 | 0.86.2 |
| `@react-native/jest-preset` | – | **0.86.2** (new) |
[...]

PR #412, opened 2026-08-31
fix(android): upgrade React Native 0.81.5 -> 0.86.3 (Part of MAGE-919)

Atharva8168:
# Description

React Native **0.81.5 → 0.86.3**. RN 0.81 is EOL, so we were building and testing against a version getting no upstream fixes.

Continues [#394](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/394) by @glenn-klaviyo — **his four commits are preserved here with him as author**. On top: a `master` merge (the branch was 9 commits behind) and 0.86.2 → 0.86.3.

**No changes under `src/`, `ios/`, or `android/src/`.** Most of the line count is `yarn.lock` and `Podfile.lock`. Note this is narrower than "no consumer-visible change" — `android/build.gradle` and `android/gradle.properties` changed, and a new `android/consumer-rules.pro` was added. **Two of those changes are consumer-visible:** the Kotlin version selection, which affects what enters a consumer's dependency graph (see the Kotlin section), and — the larger of the two — a global `-keep` rule that is now injected into every consumer's R8 run (see the AGP 9 section).

## Due Diligence

Simulator/emulator pass; **physical-device push still outstanding**. The host-version matrix below **has now been run**: React Native 0.76 through 0.87, twelve versions, every one building — debug *and* minified release — plus five versions on the iOS simulator. Running the release builds is what surfaced the AGP 9 launch crash fixed in `1371135`. Read the **Scope of what was tested** section before treating any of those tables as broad coverage.

**On the tests box — unchecked deliberately.** It was checked while this PR had no source changes at all. `android/build.gradle` has since changed materially (Kotlin version split, AGP classpath guard, explicit `kotlin-stdlib` removed), and none of that is covered by a unit test. It is not really unit-testable either: the behaviour under test is *which versions Gradle resolves in a host app we do not control*, which only a real packed install against a real template can answer. That is the host-version matrix below, and it **has now been run** across twelve React Native versions. The existing suite still covers the JS bridge surface unusually well (see the canary below), but it says nothing about the Gradle changes.

- [x] Contains readme or migration guide changes.

Public TS API unchanged. **`peerDependencies` stays `"*"`** — byte-identical to `master`, so the npm install contract is untouched and no existing install can break. README change is one line.

## Changelog / Code Overview

| Area | Change |
|---|---|
| **JS deps** | `react-native` → **0.86.3**; `react` → **19.2.8**; all `@react-native/*` → 0.86.3 (lockstep is mandatory — 0.86.3 peers `@react-native/jest-preset` at *exactly* 0.86.3) |
| **New packages** | `@react-native/jest-preset`, `@babel/core` — **consequences of RN 0.85 repackaging, not discretionary adds** |
| **Node** | `.nvmrc` → v22.23.2 because **Node 20 reached EOL on 2026-04-30** — see the correction below. Published package declares no `engines`, so contributors/CI only |
| **Android** | Kotlin version **split into standalone vs consumer selections** — see below. `kotlin-reflect` derives from whichever applies (**a latent runtime bug fix**). Explicit `kotlin-stdlib` declaration **removed**. AGP classpath guarded to standalone builds. Gradle wrapper → 9.3.1 |
| **Android (new)** | `android/consumer-rules.pro` + `consumerProguardFiles` — **one `-keep` rule that stops minified release builds crashing at launch on AGP 9.** Not a version bump; the only functional change in this PR. See the AGP 9 section |
| **iOS** | Opt out of prebuilt React-Core via `RCT_USE_PREBUILT_RNCORE` — required, verified by experiment below. Dead `fmt` `post_install` hook removed |
| **Peer range** | **Unchanged from `master`** — `react: "*"`, `react-native: "*"`. Choosing a real range is a support-policy decision, deliberately not made here |

### Verified locally

| Check | Result |
|---|---|
| `yarn install --immutable` | exit 0, no drift |
| `typecheck` / `lint` / `prepare` / `attw` | all clean |
| `yarn test` | **79 passed** (not the 57 in #394 — master added Subscription API tests) |
| example Android `assembleDebug` | **BUILD SUCCEEDED** (Gradle 9.3.1) |
| example iOS `xcodebuild` | **BUILD SUCCEEDED** |
| Gradle resolution | `react-android` 0.86.3; Klaviyo Android SDK 4.5.1; `kotlin-reflect` matches the Kotlin plugin's own stdlib |

**The canary:** `src/__tests__/index.test.tsx` hand-mocks the `react-native` module, so a shifted bridge surface breaks it with nobody touching it. It passes unchanged. Better evidence than any checklist, because it wasn't written to prove this.

**What the local checks do *not* cover:** `example/android/build.gradle` sets `ext.kotlinVersion`, so our own build always takes the `ext` branch and **never exercises the `KlaviyoReactNativeSdk_kotlinVersion` fallback**. Every green check above is blind to that path. The matrix below exists specifically to cover it.

---

> [!NOTE]
> **Correction — the Node rationale.** Earlier revisions of this description, and the commit message on `f075121`,
> said React Native 0.84 raised its Node floor to 22.11. **That was never true at any version.** Checked directly:
>
> | React Native | `engines.node` |
> |---|---|
> | 0.81.5 | `>= 20.19.4` |
> | 0.84.0 | `>= 20.19.4` — unchanged |
> | 0.85.0 | `^20.19.4 \|\| ^22.13.0 \|\| ^24.3.0 \|\| >= 25.0.0` |
> | 0.86.3 | same as 0.85 |
>
> The range narrowed at **0.85**, not 0.84, and React Native **still permits Node 20 today**. The `.nvmrc` move to
> Node 22 stands on its own reasoning — Node 20 reached end of life on 2026-04-30 — not on anything React Native
> requires. `2db6ff2` also narrows `example/package.json` `engines.node` to match RN's real range; the previous
> `>=22.11.0` admitted 22.11, 22.12 and 23.x, none of which satisfy it.

> [!IMPORTANT]
> **Correction — the AGP reasoning was wrong, and it was mine.** An earlier revision of this description, a code
> comment on `d8dcea9`, and a reply of mine on a CodeRabbit thread all claimed that `android/build.gradle` is
> published and *its buildscript classpath is consumer-visible*, so raising AGP would push that onto host apps.
>
> **That is not how Gradle loads plugins.** Plugins resolve parent-first, so a host app's own AGP declaration wins
> over anything this file declares. A host on AGP 7.2.1 stays on 7.2.1 regardless of the `8.10.1` here. This file's
> own `supportsNamespace()` already assumes exactly that — it *reads*
> `com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION` at configuration time rather than trusting the pin, which only
> makes sense if the loaded AGP can differ from the declared one.
>
> Thanks to @evan-masseau for catching it. The claim is corrected in the code comment, in this description, and on
> the CodeRabbit thread that recorded it.

> [!NOTE]
> **Known issue, pre-existing and deliberately not fixed here.** The standalone `android/gradlew` build fails at
> `:compileDebugKotlin` with `'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlin' task
> (current target is 17) jvm target compatibility should be set to the same Java version`. `compileOptions` declares
> Java 1.8 and nothing sets Kotlin's target, so Kotlin takes the JDK's.
>
> Checked against `origin/master` in a clean worktree: **it fails identically there**, with Gradle 8.14.3 and Kotlin
> 1.8.0. So the standalone build has never worked and this branch does not regress it. Aligning those targets would
> change the bytecode level our Kotlin compiles to inside consumer builds, which is out of scope for a dependency
> upgrade. Worth its own ticket, alongside the fact that nothing in CI builds `android/` as a root project — which is
> why two review findings were able to live in that `buildscript` block unexercised.

## Decisions, and why

Each of these was decided deliberately. If something below looks like an oversight, it was a choice — expand for the reasoning.

<details>
<summary><b>Why 0.86.3 and not 0.87.1 (latest)</b></summary>

0.86.3 is the current patch of the same minor this branch was built and validated against, so every iOS and Android finding here still holds. On Android that gap has since closed: 0.87.1 is measured green in the matrix below (AGP 9.2.1, Gradle 9.4.1, Kotlin 2.2.0), including minified release builds — though that only became true after the AGP 9 fix described below.

**Correction — the iOS reason no longer holds either.** An earlier revision of this section said 0.87 "would re-open the prebuilt-React-Core work against a version nothing here has validated." 0.87.1 iOS is now measured: `pod install`, a Release simulator build with prebuilt React-Core, launch, and 9/9 calls, all green on a stock scaffold (see the iOS matrix below).

What that does **not** cover is 0.87.1 against **this example app**. The 890 compile errors that motivated the prebuilt opt-out came from the example app's own dependency set (`react-native-firebase`), not from a bare scaffold — so a green scaffold at 0.87.1 says nothing about it. The remaining reason to stay on 0.86.3 is therefore scope rather than a known failure: moving to 0.87+ is **deliberately deferred, not overlooked**. Note 0.87 also raises the AGP floor, which interacts with the AGP decision below.
</details>

<details>
<summary><b>Why <code>kotlin-reflect</code> changed in a dependency PR</b> — it's a bug fix, and the failure mode is runtime</summary>

It was pinned at a hardcoded `1.8.21` while the Kotlin stdlib beside it tracked the consuming app's `$kotlin_version`. **These must resolve to the same version or reflection fails at runtime** — and this module uses reflection to enumerate Klaviyo Android SDK keywords.

Most host apps were fine *by accident*: Gradle's highest-version-wins already pulled reflect up. Apps where `1.8.21` won carried a latent failure that **ships rather than breaking the build**.

**An earlier revision of this PR took that too far.** It also moved the `KlaviyoReactNativeSdk_kotlinVersion` fallback 1.8.0 → 2.1.20, arguing that leaving it at 1.8.0 would downgrade reflect for standalone builds, and that consumers were safe because "RN templates set `ext.kotlinVersion`, and it wins when present." **The second half of that was false**, and it is the substance of @evan-masseau's review:

| RN template | Gradle wrapper | Root AGP | Sets `ext.kotlinVersion`? |
|---|---|---|---|
| 0.70 | 7.5.1 | 7.2.1 | **No** |
| 0.71 | 7.5.1 | 7.3.1 | **No** |
| 0.72 | 8.0.1 | unversioned | **No** |
| 0.73 | 8.3 | unversioned | Yes — 1.8.0 |
| 0.74 | 8.6 | unversioned | Yes — 1.9.22 |

Templates only start declaring it at **0.73**. Below that the fallback is what applies — and Kotlin plugin 2.1.20 requires **Gradle ≥ 7.6.3 and AGP ≥ 7.3.1**, both above what 0.70/0.71 ship. `README.md` claims we support 0.70+. So the bump was a real break for versions we publicly promise, not a theoretical one.

Two corrections, both in `dd9201e`:

1. **The fallback is back at 1.8.0**, matching `master`. This value now applies *only* to consumers, never to the standalone build.
2. **The explicit `kotlin-stdlib` line is gone.** `kotlin.stdlib.default.dependency` is unset, so the Kotlin plugin already injects a stdlib matching the compiler that actually runs. Declaring one ourselves instead set a floor for the whole graph — `:klaviyo-react-native-sdk:dependencies` shows requests at 1.4.10, 1.7.10, 1.8.10, 1.8.22 and 2.0.0 all resolved up to 2.1.20 — which can leave a newer stdlib under an older compiler. Removing it makes the stdlib follow the compiler automatically, which is what the reflect alignment actually needed.

3. **The Kotlin version is split in two** (`2577bff`). One property was driving two unrelated things: the plugin that runs this folder's own Gradle 9.3.1 wrapper, and the stdlib/reflect versions entering a consumer's build. Those want opposite values, so reverting to 1.8.0 fixed consumers and broke the standalone build. Now selected on `project == rootProject`, in both the buildscript classpath and the dependency block:

| Build | Kotlin version | From |
|---|---|---|
| standalone `android/gradlew` | **2.1.20** | `KlaviyoReactNativeSdk_standaloneKotlinVersion` (new) |
| host that sets `ext.kotlinVersion` | host's | `rootProject.ext.kotlinVersion` |
| host that sets none (RN ≤ 0.72) | **1.8.0** | `KlaviyoReactNativeSdk_kotlinVersion` |

2.1.20 for standalone matches upstream rather than being a guess: `@react-native/gradle-plugin@0.86.x` pins `kotlin = "2.1.20"` in its `libs.versions.toml` **and** ships a `gradle-9.3.1` wrapper, so React Native runs this pairing itself. Kotlin's published matrix stops at Gradle 8.11 for 2.1.x, which is a reason to confine the value to a build no consumer runs — not a reason to avoid it.

Verified by resolution, not assertion: standalone `buildEnvironment` loads `kotlin-gradle-plugin:2.1.20`, and `dependencies --configuration debugCompileClasspath` resolves `kotlin-reflect` and `kotlin-stdlib` to 2.1.20 — the stdlib arriving from the plugin, since we declare none.

`kotlin-reflect` still tracks `$kotlin_version`, which is the part of the original change worth keeping.
</details>

<details>
<summary><b>Why AGP stays at 8.10.1, and why its classpath is now guarded</b></summary>

**The AGP version is unchanged from `master`.** This PR does not touch it. What changed is *when it applies*.

Gradle loads plugins parent-first, and every RN app template declares AGP in its own root `build.gradle` (0.70 → 7.2.1, 0.71 → 7.3.1, 0.72+ → unversioned via the RN plugin). So inside a host app our declaration is never used. `dd9201e` guards it accordingly:

```groovy
if (project == rootProject) {
  classpath "com.android.tools.build:gradle:8.10.1"
}
```

Same pattern as `@react-native-firebase/app`, which carries this verbatim in its own `android/build.gradle`.

**The Kotlin plugin classpath is deliberately *not* guarded.** RN templates only begin declaring `kotlin-gradle-plugin` at 0.73, and a subproject's buildscript classpath is not shared with its siblings — so on 0.72 and older our line is the only thing supplying `kotlin-android` to this module. Guarding both would have broken exactly the old consumers this review set out to protect.

**Caveat worth knowing:** Gradle's compatibility matrix doesn't list AGP 8.x as tested against Gradle 9.x. That pairing now only ever occurs in a standalone build of this folder, which is neither shipped (`files` excludes `!android/gradle`, `!android/gradlew`) nor built in CI.

The Gradle wrapper bump reaches **no consumer**, for the same `files` reason, and `build.gradle` gained no Gradle-9-only syntax.
</details>

<details>
<summary><b>Why we disable prebuilt React-Core — verified by removing it</b></summary>

**Correction to the original comment:** prebuilt React-Core became default in **RN 0.84**, not 0.86. `react_native_pods.rb` forces `ENV['RCT_USE_PREBUILT_RNCORE'] = '1'` inside `use_react_native!` unless explicitly `'0'`; that line is absent in 0.83.x, present from 0.84.0. We jumped 0.81 → 0.86 in one step, so it surfaced here despite landing two minors earlier.

Rather than trust the justification, the opt-out was **removed and the app rebuilt**:

- Prebuilt React-Core comes in (122 refs in `Podfile.lock`)
- Build fails with **890 `non-modular-include-in-framework-module` errors**, all in `@react-native-firebase/app/ios/RNFBApp/*.h`

Confirmed upstream: [invertase/react-native-firebase#8883](https://github.com/invertase/react-native-firebase/issues/8883).

**Correction — the attribution above is too broad.** This section blamed the 890 errors on prebuilt React-Core becoming default in RN 0.84. The iOS matrix below contradicts the general form of that:

- On a **stock scaffold**, 0.86.3 builds cleanly **both** ways — with prebuilt React-Core, and with `RCT_USE_PREBUILT_RNCORE=0` (`logs/ios-0.86.3-src-pod.log` records "No prebuilt artifacts found, reverting to building from source" and "Removing React-Core-prebuilt"; the build is **1538 `CompileC` invocations, zero errors** — `logs/ios-0.86.3-src-build.log`).
- 0.84.1 and 0.87.1 also build green **with** prebuilt React-Core on stock scaffolds.

So the 890 errors are not a property of RN 0.84+, and not a property of prebuilt React-Core as such. They are specific to **this example app's dependency set** — most likely `@react-native-firebase/app`, whose headers every one of the 890 errors named. The opt-out remains the right call for `example/`; the generalisation was wrong.

**The `PHAsset` link error did not appear — and could not have.** The build dies at compile time, never reaching the link stage. The two symptoms are *sequential*, not parallel: the `Photos` linkage problem is only observable once the header errors are fixed. So a narrower fix (allowing non-modular includes on RNFB targets only) would likely clear 890 compile errors and then hit a link error — making the global opt-out probably correct rather than merely convenient.

Cost: slower clean iOS builds. **Example-app only** — `example/` isn't in `files` and the podspec is unchanged, so no consumer is affected.

The `fmt` hook is deleted because its target no longer exists: third-party C++ deps fold into a single `ReactNativeDependencies` pod on 0.86.
</details>

<details>
<summary><b>Why <code>peerDependencies</code> stays <code>"*"</code></b></summary>

`dependencies` is `null`, so `peerDependencies` is our entire npm contract. Since npm 7, an unsatisfiable peer is a hard `ERESOLVE` **failure**, not a warning — so tightening the range can break an existing customer's install even though nothing about their app changed. `"*"` is satisfiable by definition and therefore cannot.

An earlier revision of this PR declared `react-native: ">=0.68.0 <1.0.0"` and `react: ">=17.0.0"`. Both are gone. `05bc64e` dropped the `<1.0.0` ceiling — artificial, since RN breaks in minors, not majors. `eb4ecf1` then dropped the floor too, reverting to `master` exactly.

The reasoning for reverting rather than keeping a permissive floor: **narrowing `peerDependencies` is a breaking change, and that is a support-policy call rather than part of a version bump.** Since npm 7 an unsatisfiable peer is a hard `ERESOLVE` failure, so declaring a floor of `>=0.76` would break `npm install` outright for every consumer below it. The matrix below now measures the real range -- 0.76 through 0.87 all build -- but measuring a range and promising it are separate decisions. `README.md` separately claims 0.70+, which the matrix shows is false; that correction and the range decision are both tracked in MAGE-1176.
</details>

<details>
<summary><b>Why five RN minors produced no source changes</b> — and why that's safe, not lucky</summary>

The SDK is a **legacy module running through New-Architecture interop**. Every interop surface it uses still ships in 0.86.x, checked individually:

| Surface | 0.86.x |
|---|---|
| `ReactContextBaseJavaModule` | present |
| `ReactPackage` | present |
| `NativeModules` | present |
| `NativeEventEmitter` | present, constructor unchanged |
| `UIManager.setLayoutAnimationEnabledExperimental` | present |
| `hasActiveCatalystInstance()` | present, **deprecated** |

The checklist isn't the strongest evidence though — the canary is. `src/__tests__/index.test.tsx` hand-mocks the `react-native` module, so a shifted bridge surface breaks it without anyone touching it. It passes unchanged, and nobody wrote it to prove this point.

`hasActiveCatalystInstance()` being deprecated is exactly why the TurboModule migration is tracked as the successor (MAGE-1154): **each upgrade stays cheap only while interop survives.**
</details>

<details>
<summary><b>Why the redundant <code>-old-arch</code> CI legs are still here</b></summary>

They are genuinely redundant — agreed. On 0.86 they no longer test the legacy architecture:

- `react_native_pods.rb:118` sets `ENV["RCT_NEW_ARCH_ENABLED"] = "1"` unconditionally inside `use_react_native!`
- `@react-native/gradle-plugin`'s `ReactRootProjectPlugin.checkLegacyArchProperty` logs at `logger.error` that `newArchEnabled=false` "is not supported anymore since React Native 0.82"

Both legs build the New Architecture, costing ~21 min per run plus a duplicate cache set, while showing a green check implying coverage we don't have.

**Removing them isn't just deleting YAML.** Four of those job names are required status checks on `master`, and there's a code half — `android/build.gradle` skips `apply plugin: "com.facebook.react"` when the flag is false, still reachable because RN only warns. Kept out so this PR stays a dependency upgrade. Tracked as MAGE-1155.
</details>

<details>
<summary><b>Merge conflict resolutions</b> — three files, nothing lost from master</summary>

| File | Resolution |
|---|---|
| `android/gradle.properties` | `klaviyoAndroidSdkVersion=4.5.1` (master) **+** `kotlinVersion` (branch). Conflicted only as adjacent lines. The `kotlinVersion` value has since been reverted to master's `1.8.0`, and a separate `standaloneKotlinVersion` added — see the Kotlin section |
| `package.json` | master wins `version` 2.5.0 and `protobufjs ^7.6.5` (security bump); branch wins `@types/react ^19.2.0` (required by RN 0.86) |
| `example/ios/Podfile.lock` | **Regenerated, not hand-merged** — CocoaPods refused outright, the stale lockfile pinned 0.86.2's local podspecs |

`yarn.lock` auto-merged textually, but a textual auto-merge of a lockfile isn't a resolved lockfile — regenerated and re-verified with `--immutable`.

**Not included on purpose:** no CI changes (the redundant `-old-arch` matrix cleanup is tracked separately), and no `example/Gemfile.lock` change (local bundler wanted `minitest` 5→6; reverted).
</details>

<details>
<summary><b>Process: why merge rather than rebase, and who authored what</b></summary>

**Integrated by merge, not rebase**, because the upstream branch was shared — none of @glenn-klaviyo's history is rewritten. The trade-off is one merge commit in the history rather than a linear four.

**Authorship:** opened by @glenn-klaviyo as #394, carried to completion here. His four commits are intact with him as author — `git log` shows this. #394 was closed in favour of a single review surface rather than a two-hop merge through his branch, since `git` preserves the credit either way.

**Three things found in review of this PR and fixed here**, all worth knowing because they were wrong in the original write-up:
- The Podfile comment attributed prebuilt React-Core to RN 0.86. It was **0.84** (`d8dcea9`).
- `@react-native/jest-preset` was referenced by `example/jest.config.js` but declared only at the root — a regression this branch introduced, since master's `preset: 'react-native'` *was* declared in the example workspace (`390bf98`).
- The AGP consumer-visibility claim was false, and the Kotlin fallback bump that rested on it was a real break for RN ≤ 0.72 (`eb4ecf1`, `dd9201e`). See the note above.
</details>

---

## Still to do

- [x] **Host-version matrix.** Run across **twelve** React Native versions, 0.76 through 0.87 — see the table below.
- [ ] **Physical-device push**: token registration and delivery on real iOS and Android hardware. Not obtainable on simulator/emulator — the iOS Simulator can't get a genuine APNs token and FCM needs Play Services.
- [x] CI green on this branch.
- [x] **Emulator runtime check (debug builds)** on 0.76.9 / 0.86.3 / 0.87.1 — see below.
- [x] **Minified release builds**, all twelve versions, launched and exercised — see below. This is what surfaced the AGP 9 launch crash and the fix in `1371135`.
- [x] **iOS matrix** — five versions, Release simulator builds — see below.

## Host-version matrix

`klaviyo-react-native-sdk-2.5.1.tgz` packed from this branch and installed into a stock
React Native app per version, scaffolded with
`@react-native-community/cli@latest init --version <X>`. `app builds` is
`:app:assembleDebug`, which runs `checkDebugAarMetadata` — the gate that actually rejects
this SDK on older React Native. JDK 17.0.19, Node 26.4.0.

| RN | AGP (host) | AGP (we declare) | Kotlin plugin | kotlin-stdlib | kotlin-reflect | module compiles | app builds |
|---|---|---|---|---|---|---|---|
| 0.76.9 | 8.6.0 | none (guarded) | 1.9.25 | 2.0.0 | 1.9.25 | pass | **pass** |
| 0.77.3 | 8.7.2 | none (guarded) | 2.0.21 | 2.0.21 | 2.0.21 | pass | **pass** |
| 0.78.3 | 8.8.0 | none (guarded) | 2.0.21 | 2.0.21 | 2.0.21 | pass | **pass** |
| 0.79.7 | 8.8.2 | none (guarded) | 2.0.21 | 2.0.21 | 2.0.21 | pass | **pass** |
| 0.80.3 | 8.9.2 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.81.5 | 8.11.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.82.1 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.83.10 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.84.1 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.85.3 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.86.3 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass | **pass** |
| 0.87.1 | 9.2.1 | none (guarded) | 2.2.0 | 2.2.10 | 2.2.0 | pass | **pass** |

**All twelve build.** Four things this establishes:

1. **`AGP (we declare)` is `none (guarded)` on every row.** Plugins load parent-first, so the
   host's AGP always wins; ours never reaches a consumer. This is the claim an earlier
   revision of this PR got backwards, now measured rather than argued.
2. **The `KlaviyoReactNativeSdk_kotlinVersion=1.8.0` fallback never fires in this range.**
   Every template from 0.76 up sets `ext.kotlinVersion`, so the `rootProject.ext` branch is
   always taken. The fallback only applies to 0.72 and older.
3. **Two rows have a `kotlin-stdlib` / `kotlin-reflect` skew**, neither fatal to the build:
   - **0.76.9** — reflect 1.9.25 against stdlib 2.0.0. The template's KGP is 1.9.25, but the
     Klaviyo Android SDK drags stdlib to 2.0.0 via coroutines 1.9.0. Reflect is held at 1.9.25
     by a strict constraint, so it cannot be raised from here.
   - **0.87.1** — reflect 2.2.0 against stdlib 2.2.10, a patch-level gap: AGP 9.2.1 depends on
     `kotlin-gradle-plugin 2.2.10` while the template sets `ext.kotlinVersion = "2.2.0"`.

   Both are the tolerant direction (older reflect, newer stdlib) and neither fails the build,
   but reflection paths are untested against them.
4. **0.87.1 passes on AGP 9** without the fix in #416, because RN 0.87's template ships
   `android.builtInKotlin=false`. That guard is still correct defense for a consumer who flips
   the flag, but nothing in the stock 0.87 template requires it.

**Below 0.76 remains broken**, and not because of anything in this PR:
`klaviyo-android-sdk:analytics` pins `androidx.core:core-ktx:{require 1.16.0; reject _}`, whose
AAR metadata demands AGP 8.6.0. RN 0.76 is the first template shipping exactly that. Raising
that floor is Android-SDK-side work, tracked separately.

## Emulator runtime check

The matrix above proves twelve versions *compile*. It cannot show the SDK *runs* — and the
two skew rows are runtime risks specifically. So three versions were run on an emulator
(Pixel_10, API 36): both skew rows plus the shipping target.

A stock React Native app never imports this SDK, so the native module is never constructed
and `getConstants()` never fires. Each app's `App.tsx` was given a touch of
`NativeModules.KlaviyoReactNativeSdk` to force construction, which is what triggers the
reflection in `extractConstants()`.

| RN | Kotlin config | PROFILE_KEYS | EVENT_NAMES | Crashes |
|---|---|---|---|---|
| 0.76.9 | **skew** — reflect 1.9.25 / stdlib 2.0.0 | 16 | 4 | 0 |
| 0.86.3 | matched 2.1.20 | 16 | 4 | 0 |
| 0.87.1 | **skew** — reflect 2.2.0 / stdlib 2.2.10, AGP 9 | 16 | 4 | 0 |

```
ReactNativeJS: 'KLAVIYO_SMOKE_OK', '["FORMS_AVAILABLE","LOCATION_AVAILABLE","EVENT_NAMES","PROFILE_KEYS"]'
ReactNativeJS: 'KLAVIYO_SMOKE_PROFILE_KEYS', '["location","TITLE","REGION","LAST_NAME","ZIP","FIRST_NAME",
  "IMAGE","LONGITUDE","COUNTRY","CITY","ADDRESS2","ORGANIZATION","properties","TIMEZONE","LATITUDE","ADDRESS1"]'
ReactNativeJS: 'KLAVIYO_SMOKE_EVENT_NAMES', '["VIEWED_PRODUCT","STARTED_CHECKOUT","OPENED_APP","ADDED_TO_CART"]'
```

Identical key sets on all three. Zero `FATAL EXCEPTION`, `KotlinReflectionInternalError`,
`NoSuchMethodError`, or `ClassNotFoundException` in any log.

**Both skews are harmless in practice.** `T::class.nestedClasses` -> `visibility` ->
`objectInstance` — the entire reflection surface this SDK uses — resolves correctly with
`kotlin-reflect` one minor behind `kotlin-stdlib` (0.76) and one patch behind (0.87).
Non-empty results were the bar, since a silent empty list was the realistic failure mode
rather than a crash.

0.76.9 ran with `Bridgeless mode is enabled` and `fabric:true`, so the legacy module went
through New-Architecture interop — the real-world path, not a legacy-arch fallback.

**One environmental note:** on 0.87.1 the first launch did not render. Android 16 /
targetSdk 36 raised an automatic `ACCESS_LOCAL_NETWORK` permission dialog that took focus,
Metro's connection timed out, and the Fabric surface never started. Granting permissions and
relaunching fixed it. That is a dev-menu artifact rather than SDK behaviour, but customers
targeting SDK 36 on 0.87 may meet the same dialog.

Everything above is a **debug** build. Release builds with R8 are covered separately below — they were not, at the time this section was written, and running them is what turned up the AGP 9 launch crash. **Push delivery is still genuinely not covered**: it needs hardware.

## The AGP 9 release-build crash, and the one functional change in this PR

`1371135` adds `android/consumer-rules.pro` and wires it in with `consumerProguardFiles "consumer-rules.pro"` in `android/build.gradle`. It is the only change here that is not a version bump, and it is a workaround for a bug that is not ours.

**What happens.** RN 0.87's template ships **AGP 9.2.1**, whose R8 removes the no-arg constructor from Room's generated `androidx.work.impl.WorkDatabase_Impl` while keeping the class itself. WorkManager builds that database from `androidx.startup`'s ContentProvider, which Android starts *before* `Application.onCreate` — so a minified release build dies at launch with no app code having run:

```
java.lang.RuntimeException: Unable to get provider androidx.startup.InitializationProvider
Caused by: Failed to create an instance of class androidx.work.impl.WorkDatabase
  at androidx.work.WorkManagerInitializer
```

The constructor is what goes, not the class. Two things pin that down: there is **no `ClassNotFoundException`** anywhere in the crashing logs, and a keep rule naming only `<init>()` fixes it.

**How it reaches us.**

```
klaviyo-android-sdk:analytics:4.5.1
  -> androidx.work:work-runtime-ktx:{require 2.9.0; reject _}
```

Read straight out of the published Gradle module metadata for `analytics:4.5.1`. `work-runtime` 2.9.0 predates AGP 9, and its own `proguard.txt` keeps Workers but says nothing about the Room database. Because the pin is **strict**, a consuming app cannot resolve its way out of it.

**It is not introduced by this PR.** Two independent reproductions:

- The **published `klaviyo-react-native-sdk@2.5.0`**, installed from npm into an RN 0.87.1 app, crashes identically — `logs/r8-build-0.87.1-npm250.log`, `logs/r8-0.87.1-npm250.log`.
- A **plain native Kotlin app with no React Native in it at all** — AGP 9.2.1, Gradle 9.4.1, compileSdk 36, one dependency (`com.github.klaviyo.klaviyo-android-sdk:analytics:4.5.1`), `Klaviyo.initialize()` never called — crashes the same way in release (`native-agp9/logcat-release.log`) and launches fine in debug (`native-agp9/logcat-debug.log`).

**Isolation.**

| Ruled out | Evidence |
|---|---|
| The ProGuard config | 0.86.3 passes with the same aggressive `proguard-android-optimize.txt` — `logs/r8-build-0.86.3-optimize.log`, `logs/r8-0.86.3-optimize.log` |
| The `kotlin-stdlib` / `kotlin-reflect` skew | removing it changes nothing; still crashes — `logs/r8-0.87.1-kotlin2210.log` |
| This branch | the published 2.5.0 crashes identically |
| Minification as such | debug builds are unaffected, and every AGP 8 row minifies without incident |

> [!NOTE]
> **Provenance of the first two rows.** Their *outcomes* are in the logs — `BUILD SUCCESSFUL`
> for the optimize run, `FATAL EXCEPTION` for the matched-Kotlin run. Their *configuration* is
> not: Gradle does not record which `getDefaultProguardFile(...)` or which `ext.kotlinVersion`
> a build used, and each run's edit was overwritten by the next build. So "it used the optimize
> file" and "it used Kotlin 2.2.10" rest on run provenance rather than on anything inside the
> artifacts. Both are believed correct, and neither is load-bearing: the cause and the fix are
> established by the three rows that *are* independently checkable — the published-2.5.0 crash,
> the native-app-without-React-Native crash, and the keep rule fixing both.

**The fix** — `-keep class androidx.work.impl.WorkDatabase_Impl { <init>(); }` — is verified against **both** the published 2.5.0 (`logs/r8-0.87.1-npm250-keeprule.log`) and this branch (`logs/r8-0.87.1-consumerrule2.log`, `logs/r8-0.87.1-narrow.log`). It genuinely travels through the AAR: the rule appears in a *consuming* app's merged R8 configuration at `App0871/android/app/build/outputs/mapping/release/configuration.txt`.

> [!WARNING]
> **One cell is void, and is not being claimed.** An attempt to test 0.87.1 with the plain `proguard-android.txt` produced a crash, but that run proves nothing: the build never configured — `BUILD FAILED in 3s`, `logs/r8-build-0.87.1-plain.log` — because AGP 9 rejects `getDefaultProguardFile('proguard-android.txt')` outright for containing `-dontoptimize`. The crash observed alongside it came from a stale APK. **Do not read "0.87.1 with plain ProGuard also crashes" anywhere in this PR.**

**Scope — read this before assuming the bug is fixed.** The rule belongs in `klaviyo-android-sdk`, not in this wrapper. It is tracked there as **MAGE-1199**. What ships here covers **React Native consumers only**. Native Android consumers of `klaviyo-android-sdk:analytics` on AGP 9 hit the identical crash and are **not** covered — that is exactly what the native reproduction above demonstrates. `consumer-rules.pro` documents its own removal condition.

## Android release/API matrix

All twelve versions rebuilt with `minifyEnabled true`, installed as release APKs, launched, and driven through nine `Klaviyo` calls — each wrapped in a JS `try`/`catch`. Same Pixel_10 / API 36 emulator. Source of truth: `logs/api-matrix-results.txt` and `logs/api-*.log`; driver script `api-matrix.sh`.

| RN | release build | install + launch | calls | JS-visible failures | `FATAL EXCEPTION` |
|---|---|---|---|---|---|
| 0.76.9 | pass | pass | 9 | 0 | 0 |
| 0.77.3 | pass | pass | 9 | 0 | 0 |
| 0.78.3 | pass | pass | 9 | 0 | 0 |
| 0.79.7 | pass | pass | 9 | 0 | 0 |
| 0.80.3 | pass | pass | 9 | 0 | 0 |
| 0.81.5 | pass | pass | 9 | 0 | 0 |
| 0.82.1 | pass | pass | 9 | 0 | 0 |
| 0.83.10 | pass | pass | 9 | 0 | 0 |
| 0.84.1 | pass | pass | 9 | 0 | 0 |
| 0.85.3 | pass | pass | 9 | 0 | 0 |
| 0.86.3 | pass | pass | 9 | 0 | 0 |
| 0.87.1 | pass | pass | 9 | 0 | 0 |

The calls, in order: `setLoggingEnabled`, `initialize`, `setEmail`, `setPhoneNumber`, `setExternalId`, `setProfileAttribute`, `setProfile`, `createEvent`, `resetProfile`.

**Two caveats, and the table is misleading without them.**

**(a) Eleven of the twelve rows ran with R8's optimizer switched OFF.** Only **App0871 (0.87.1)** uses `getDefaultProguardFile("proguard-android-optimize.txt")`. The other eleven use `getDefaultProguardFile("proguard-android.txt")`, which contains **`-dontoptimize`** — verified by grepping every `App*/android/app/build.gradle`, and again downstream in the merged configs (App0863's `configuration.txt` carries `-dontoptimize`; App0871's does not).

So the honest reading is **one version tested under full R8 and eleven under shrink + obfuscate only** — and the optimization pass is precisely what produces the WorkManager bug documented above. 0.87.1 is on the optimize file because AGP 9 refuses the other one, not because the matrix was designed that way.

**(b) "9 calls, zero failures" is a weaker claim than it looks.** It means the bridge accepted nine calls without throwing **synchronously**, and the process stayed alive. Legacy RN bridge methods are async fire-and-forget, so a native-side exception cannot reach the JS `try`/`catch` that produced these counts.

And nothing in these logs shows the native SDK doing anything. `setLoggingEnabled(true)` is the **first** call, yet across all twelve Android logs there are **zero** Klaviyo SDK log tags, **zero** references to `api.klaviyo.com`, and **zero** occurrences of the API key. The key is fake (`TESTKY`). **No claim is made here that any request was formed or sent.**

What the table does establish — and it is the thing that was broken a day ago — is that a minified release APK **builds, installs, launches, and survives the call sequence** on all twelve versions.

## iOS matrix

All twelve versions, Release simulator builds (`-configuration Release`, Xcode 26.6 / build 17F113), same nine-call test. Source: `logs/ios-*-pod.log`, `logs/ios-*-build.log`, `logs/ios-*.log`.

| RN | `fmt` | `pod install` | build | launch | calls |
|---|---|---|---|---|---|
| 0.76.9 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.77.3 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.78.3 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.79.7 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.80.3 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.81.5 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.82.1 | 11.0.2 | pass | **FAIL — not our SDK** | n/a | n/a |
| 0.83.10 | 12.1.0 | pass | pass (from source) | pass | 9/9 |
| 0.84.1 | prebuilt | pass | pass (prebuilt React-Core) | pass | 9/9 |
| 0.85.3 | 12.1.0 | pass | pass (from source) | pass | 9/9 |
| 0.86.3 | prebuilt | pass | pass (prebuilt **and** from source) | pass | 9/9 |
| 0.87.1 | 12.1.0 | pass | pass (prebuilt React-Core) | pass | 9/9 |

`fmt` reads `prebuilt` where React-Core ships as an xcframework that bundles it rather than
installing it as a separate pod.

**Every failure is the same Xcode problem, and none of them are ours.** RN 0.76.9 through
0.82.1 vendor `fmt` 11.0.2, which does not compile under Xcode 26.6's clang: the build aborts in
`Pods/fmt/src/format.cc` at `format-inl.h:59:24`, `error: call to consteval function ... is not a
constant expression`, before React-Core is even reached. React Native fixed this in 0.83 by moving
to `fmt` 12.1.0 — the boundary is exactly between 0.82.1 and 0.83.10, with nothing gradual about it.

For **all seven** failures a control app was scaffolded at the same RN version **with the Klaviyo
SDK removed from `package.json`** (`grep -ci klaviyo Podfile.lock` returns 0 in every one). All
seven controls fail with the identical error at the identical line — `logs/ios-<version>-control-build.log`.
Not once did a control build while ours failed.

Two consequences worth stating separately:

- This does **not** lower our iOS support floor. It means RN 0.82 and older cannot be built by
  Xcode 26.6 at all, Klaviyo or otherwise. A customer on an Xcode contemporary with their RN
  version is unaffected, and our SDK may well work there — untested either way.
- It **does** cap what is verifiable here. No claim is made about iOS on RN 0.76-0.82; those rows
  are untestable on this toolchain rather than known-bad.

Combined with Android, the floors differ by platform: **Android 0.76** (our `androidx.core` pin),
**iOS 0.83** (React Native's `fmt`, on this Xcode). A cross-platform app is bound by the higher of
the two.

**The same async caveat as (b) above applies to the 9/9 column** — those counts come from JS `console.log` lines. The iOS evidence is nonetheless somewhat stronger than the Android evidence, and not because the test does anything different: on iOS the native SDK actually logs. `logs/ios-0.86.3.log` contains

```
[com.klaviyo.klaviyo-swift-sdk.klaviyoCore:Event Buffering] 📤 Buffering event: OPENED_APP
[com.klaviyo.klaviyo-swift-sdk.klaviyoCore:Event Buffering] 💾 Buffer now has 1 event(s)
```

plus a DNS resolution for `a.klaviyo.com` — so `createEvent` demonstrably crossed the bridge and reached native code. Nothing shows a request completing or being accepted; the API key is fake. The Android logs show no equivalent native-side activity at all.

## Scope of what was tested

Stated plainly, because the tables above are easy to over-read:

- **One patch per RN minor.** 0.76.9 stands in for all of 0.76.x, 0.86.3 for all of 0.86.x, and so on. Nothing here says anything about the other patch releases in each minor.
- **Stock scaffolds only.** Every matrix app is a bare `@react-native-community/cli init` project with **no other native modules**. The one run against a realistic dependency set is the example app, at **0.86.3 alone**.
- **Nothing ever reached Klaviyo.** Fake API key throughout, so no event was delivered. Push was never tested — there is no `google-services.json` in this repo and real delivery needs hardware. In-app forms never rendered. Geofencing / location never exercised.
- **One machine, one emulator image, one toolchain.** A single Pixel_10 / API 36 AVD, JDK 17.0.19, Node **26.4.0** — note the matrix did **not** run on this branch's own `.nvmrc` version (22.23.2).
- **CI still builds one React Native version.** There is no version matrix in `.github/workflows`; the twelve above were run by hand and will not re-run on the next PR.

## Follow-ups filed

MAGE-1154 (TurboModule migration, with sub-issues MAGE-1158/1159/1160) · MAGE-1155 (redundant `-old-arch` CI legs) · MAGE-1156 (customer docs for the RNFB iOS build failure, **0.84+**) · MAGE-1157 (`MainActivity` cold-start log noise) · **MAGE-1199 (`androidx.work` strict pin → AGP 9 startup crash; the durable fix for the keep rule shipped here, and the ticket that must land before native Android consumers are covered)** · MAGE-1176 (`README.md`'s `0.70+` claim and the `peerDependencies` range decision)

## A note on the size of this PR

Apologies for the length — ~1,800 changed lines is more than anyone wants to open. Splitting was considered and deliberately rejected:

- **Most of the diff isn't hand-written.** `yarn.lock` and `example/ios/Podfile.lock` are the overwhelming majority of those lines. The hand-edited part is roughly a hundred lines across eight files.
- **The obvious split points produce PRs that fail CI.** Bumping `react-native` without the iOS Podfile change pulls in prebuilt React-Core, which fails with 890 compile errors (see the iOS section above). The Podfile change is meaningless before the version bump. Splitting would land non-building commits on `master` and break `git bisect` through that range.
- **CI cost multiplies.** Every run carries **~21 minutes of redundant compute** from the duplicated `-old-arch` legs (MAGE-1155). Four PRs would pay that four times over for no added safety.

If you'd like the staged reading a split would have given: **the commits are already separated by concern** — deps, Android, iOS, docs — so reviewing commit-by-commit gets you there without the broken intermediate states.

## Related Issues/Tickets

Closes MAGE-919

## Summary by CodeRabbit

* **Updates**
  * Upgraded the example app and development environment to React Native 0.86.3, React 19.2.8, Node.js 22, Kotlin 2.1.20, and Gradle 9.3.1.
  * Improved compatibility with React Native’s new architecture across Android and iOS.
  * Refined Android and iOS build configuration for more reliable builds, headers, modules, and Photos integration.
* **Documentation**
  * Updated React Native version and new-architecture compatibility guidance.
* **Testing**
  * Updated Jest configuration to use the React Native testing preset.


android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,8 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
```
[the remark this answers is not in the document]
Atharva8168: Fixed in 2f269fd — added `distributionSha256Sum` to both wrappers, using Gradle's published checksum for `gradle-9.3.1-bin.zip` (`b266d5ff...ff06`).

Verified by deleting `~/.gradle/wrapper/dists/gradle-9.3.1` and re-running both wrappers: each re-downloaded, passed verification, and reported Gradle 9.3.1.

Worth noting the gap was pre-existing — `master` has the same `validateDistributionUrl=true` and no checksum. Fixing it here because this PR changes the `distributionUrl` line, so hardening that same download belongs with it.

android/gradle.properties
```diff
@@ -18,7 +18,7 @@ android.useAndroidX=true
 # org.gradle.parallel=true
 #Tue Dec 19 15:08:27 EST 2023
 KlaviyoReactNativeSdk_klaviyoAndroidSdkVersion=4.5.1
-KlaviyoReactNativeSdk_kotlinVersion=1.8.0
+KlaviyoReactNativeSdk_kotlinVersion=2.1.20
```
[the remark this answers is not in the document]
Atharva8168: Not changing this one — the combination is React Native's own, not ours.

`@react-native/gradle-plugin@0.86.x` pins Kotlin **2.1.20** in its `gradle/libs.versions.toml` and ships Gradle **9.3.1** in its own wrapper. We match upstream exactly; we are not picking an unusual tuple.

So the compatibility-table objection is real, but it lands on React Native rather than on this SDK. Both suggested remedies make things worse for consumers:

- **Kotlin 2.2.0+** — diverges from what RN's own plugin is compiled against, and `KlaviyoReactNativeSdk_kotlinVersion` is a *published* consumer-facing fallback. Trades a theoretical mismatch for a real one.
- **Gradle ≤ 8.11** — below what the RN 0.86 plugin ships, likely to break it outright.

Scope note: this value is a fallback. A consuming app that sets `rootProject.ext.kotlinVersion` overrides it, so it only applies to consumers who set nothing and to standalone library builds.

Evidence it works today: both Android CI legs pass, and Gradle resolves `kotlin-reflect` and `kotlin-stdlib` to a matching 2.1.20 against the example app.

AGP is separately held at 8.10.1 on purpose — `android/build.gradle` is published and its buildscript classpath is consumer-visible, so raising it pushes that onto host apps. There is a comment beside the pin explaining this.

evan-masseau (commented):
As of 0.82 the old architecture is no longer supported, could you analyze whether the CI workflow matrix of old/new arch is worth maintaining anymore? See what alternatives / improvements to our build verification we might consider building instead, whether for this PR or a follow up. 

PS, unless we're _certain_ this feature is all-in-one branch (it may not be, e.g. if we need to do a separate PR for some CI workflow improvements) then I think you should make a branch off master like `feat/react-native-86` and target that one for PRs for now.

package.json
```diff
@@ -98,19 +100,19 @@
     "typescript": "^5.8.3"
   },
   "resolutions": {
-    "@types/react": "^19.1.0",
+    "@types/react": "^19.2.0",
     "protobufjs": "^7.6.5"
   },
   "peerDependencies": {
-    "react": "*",
-    "react-native": "*"
```
evan-masseau, line 106: Why eliminate the wildcards? Does our readme advertise a specific range of versions already, and is this within that? Also why `<1.0.0` if we don't know what `1.0` will bring, that seems like an artificial ceiling

Atharva8168:
Good call @evan-masseau  and I feel you are right, and the analysis backs you up.

React Native 0.86 turns the New Architecture on by itself (react_native_pods.rb:118), and the Gradle plugin logs an error that newArchEnabled is false is not supported since 0.82. So both -old-arch legs build the New Architecture. Cost is 5m24s + 15m51s per run. They also keep a second set of caches, which pushes out the caches the real jobs need. I tracked in MAGE-1155.

Two things to add ->

Timing. master still runs 0.81, where the setting does work. So today those legs do test something real. Merge the upgrade first, then remove them.

Your second question is the better one, and I had not asked it. I only looked at what to remove. You asked what to build with the freed time. Four ideas, all follow-up:

1. Test several React Native versions instead of two architectures. We promise >=0.68.0 and test one.
2. Run pack-and-test.sh on pull requests. It exists, but runs only at publish time. This overlaps with your MAGE-1075.
3. Build android/ on its own. That would expose the jvmTarget mismatch, which is invisible today.
4. Build a release variant. Everything today is Debug.

I would keep all four out of this PR, so it stays a dependency change with no SDK source edits. Happy to file them wherever you think they fit best.

Atharva8168:
Also 100% agreed to the branching suggestion, will do the necessary changes. @evan-masseau

Atharva8168:
DUPLICATED Message already there in the thread above

My bad on the ceiling — you are right, that was my call and it was wrong. Dropping it.

Wildcards. MAGE-919 asks for it directly. With * ,  npm can never warn anyone about an unsupported version.

README range. Yes, and the floor comes from there. Requirements lists 0.68.x–0.72.x as compiling "with the following modifications", and 0.74+ as "fully tested and verified". So >=0.68.0 matches our own docs.

<1.0.0. Artificial, agreed 100%. React Native breaks in minors, 0.82 removed the old architecture — so a 1.0 cap guards against nothing that has ever broken us. It also contradicts my own argument in this PR for leaving react uncapped.
Changing to react-native: ">=0.68.0", matching react: ">=17.0.0".

Atharva8168:
Peer ceiling. Dropped <1.0.0, so the range is now react-native: ">=0.68.0" and matches react: ">=17.0.0". yarn.lock needed regenerating too — Yarn records the peer range there, and --immutable failed until I updated it. (05bc64e)

Branch. Cut feat/react-native-86 from master and retargeted this PR to it.
 
CI matrix. Already ticketed — I filed MAGE-1155 yesterday
  
pack-and-test.sh. That one overlaps your MAGE-1075 — both touch publish-example.yml. It probably belongs in your Continuous Deployment milestone rather than a new ticket. The other three ideas are unfiled. Tell me where you want them.

package.json
Atharva8168: My bad on the ceiling — you are right, that was my call and it was wrong. Dropped in `05bc64e`.

**Wildcards.** MAGE-919 asks for it directly. With `*`, npm can never warn anyone about an unsupported version.

**README range.** Yes, and the floor comes from there. Requirements lists 0.68.x–0.72.x as compiling "with the following modifications", and 0.74+ as "fully tested and verified". So `>=0.68.0` matches our own docs rather than being a number I picked.

**`<1.0.0`.** Artificial, agreed. React Native breaks in minors — 0.82 removed the old architecture — so a 1.0 cap guards against nothing that has ever broken us. It also contradicts my own argument in this PR for leaving `react` uncapped.

Now `react-native: ">=0.68.0"`, matching `react: ">=17.0.0"`. `yarn.lock` needed regenerating too — Yarn records the peer range there, and `--immutable` failed until I updated it.

(Posted this in the main conversation earlier by mistake — repeating it here so the thread has it.)

android/gradle.properties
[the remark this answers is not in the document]
Atharva8168: Reopening this to correct my own reply above, because part of it was wrong and it ended up in a stored learning.

**The AGP claim was false.** I wrote that `android/build.gradle` is published and "its buildscript classpath is consumer-visible, so raising it would affect host applications." Gradle loads plugins **parent-first**, so a host app's own AGP declaration wins over anything this file declares. A host on AGP 7.2.1 stays on 7.2.1 no matter what `8.10.1` says here. This same file already assumes that: `supportsNamespace()` reads `com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION` at configuration time instead of trusting the pin, which only makes sense if the loaded AGP can differ from the declared one.

The stored learning from this thread repeats that claim, so please drop it.

**The Kotlin value also changed.** I said we were not changing `KlaviyoReactNativeSdk_kotlinVersion=2.1.20` because it matches `@react-native/gradle-plugin@0.86.x`. That reasoning was incomplete. The property is only read when a host sets no `rootProject.ext.kotlinVersion`, and React Native templates only start setting it at **0.73**:

| RN template | Gradle wrapper | Root AGP | Sets `ext.kotlinVersion`? |
|---|---|---|---|
| 0.70 | 7.5.1 | 7.2.1 | No |
| 0.71 | 7.5.1 | 7.3.1 | No |
| 0.72 | 8.0.1 | unversioned | No |
| 0.73 | 8.3 | unversioned | Yes — 1.8.0 |

Kotlin plugin 2.1.20 requires **Gradle ≥ 7.6.3 and AGP ≥ 7.3.1**. RN 0.70 and 0.71 ship Gradle 7.5.1 and AGP 7.2.1, both below that floor, and our README claims support for 0.70+. So the original finding on this line was closer to right than my reply allowed — it just landed on the wrong remedy.

It is now reverted to `1.8.0`, matching `master`. The explicit `kotlin-stdlib` declaration is removed as well, since `kotlin.stdlib.default.dependency` is unset and the Kotlin plugin already injects a stdlib matching the compiler that runs. `kotlin-reflect` still tracks `$kotlin_version`.

For the record, so a corrected learning can replace the old one:

- The AGP version in `android/build.gradle` is **not** consumer-visible; plugin resolution is parent-first and the host wins.
- Its classpath entry is now guarded with `if (project == rootProject)` so it only applies to standalone builds of the `android/` folder.
- The `kotlin-gradle-plugin` classpath is deliberately **not** guarded: RN templates only declare it from 0.73, and a subproject's buildscript classpath is not shared with siblings, so below 0.73 our entry is the only thing supplying `kotlin-android` to this module.
- `KlaviyoReactNativeSdk_kotlinVersion` is `1.8.0`, chosen to stay within RN 0.70's Gradle 7.5.1, not to match the RN 0.86 toolchain.

package.json
Atharva8168: Following up here, because this thread now says something I have since changed.

I wrote that the range was `react-native: ">=0.68.0"`. It is not. It is back to `"*"` for both `react` and `react-native`, matching `master` exactly — `eb4ecf1`.

You were right twice, not once. Dropping the `<1.0.0` ceiling was the first fix. But the floor had the same problem in a smaller form: a range is a promise, and we have never measured which versions we can keep. The README says 0.70+ and notes 0.68.7 as compiling, but nothing in CI tests any of it. `>=0.68.0` was me reading our own docs back at you, not evidence.

So the install contract is untouched by this PR. Picking a real range needs the host-version matrix behind it, and that is a support-policy call rather than part of a version bump.

Atharva8168:
Thanks for this — all three land, and one of them was my mistake rather than a difference of opinion. Taking them in order.

**1. Peer dependencies.** Agreed, and reverted further than you asked. Both `react` and `react-native` are back to `"*"`, byte-identical to `master` (`eb4ecf1`). The `<1.0.0` ceiling went first, but the floor had the same flaw: a range is a promise, and we have never measured which versions we can keep. Our README claims 0.70+ and notes 0.68.7 as compiling, and nothing in CI tests any of it. So the install contract is untouched by this PR. I have amended MAGE-919 to drop "peer range declared" from its acceptance criteria, since it can't quietly disagree with what ships.

**2. AGP and parent-first loading.** **My bad — I will call this one on myself.** I claimed this file's buildscript classpath is consumer-visible and that raising AGP would push it onto host apps. That is simply not how Gradle loads plugins, and your AGP 7.2.1 example proves it. Worse, this file already assumed you were right: `supportsNamespace()` reads `com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION` at configuration time rather than trusting the pin, which only makes sense if the loaded AGP can differ from the declared one. I wrote the claim into a code comment, the PR description, and a CodeRabbit reply. All three are corrected, and I have asked CodeRabbit to drop the stored learning that repeated it.

The AGP entry is now guarded so it only applies when this folder is its own root project:

```groovy
if (project == rootProject) {
  classpath "com.android.tools.build:gradle:8.10.1"
}
```

Same pattern `@react-native-firebase/app` uses.

**One deviation from your suggestion, and I want to flag it rather than bury it.** I did not guard the Kotlin plugin the same way. RN templates only begin declaring `kotlin-gradle-plugin` at **0.73**, and a subproject's buildscript classpath is not shared with its siblings — so on 0.72 and older our entry is the only thing supplying `kotlin-android` to this module. Guarding both would have broken exactly the old consumers we are trying to protect. Tell me if you read that differently.

**3. Splitting the Kotlin version.** Agreed, and this was the real break. Looking at the templates:

| RN template | Gradle wrapper | Root AGP | Sets `ext.kotlinVersion`? |
|---|---|---|---|
| 0.70 | 7.5.1 | 7.2.1 | **No** |
| 0.71 | 7.5.1 | 7.3.1 | **No** |
| 0.72 | 8.0.1 | unversioned | **No** |
| 0.73 | 8.3 | unversioned | Yes — 1.8.0 |
| 0.74 | 8.6 | unversioned | Yes — 1.9.22 |

You were right that 0.73 is the boundary. Kotlin plugin 2.1.20 needs **Gradle ≥ 7.6.3 and AGP ≥ 7.3.1**, and 0.70/0.71 ship 7.5.1 and 7.2.1 — both below it. So moving the fallback to 2.1.20 broke versions we publicly claim to support, and I had defended it on the false premise that templates always set `ext.kotlinVersion`.

Two fixes in `dd9201e`:

- **Fallback back to `1.8.0`**, matching `master`.
- **Explicit `kotlin-stdlib` declaration removed.** `kotlin.stdlib.default.dependency` is unset, so the Kotlin plugin already injects a stdlib matching the compiler that actually runs. Ours was instead setting a floor for the whole graph — `:klaviyo-react-native-sdk:dependencies` shows requests at 1.4.10, 1.7.10, 1.8.10, 1.8.22 and 2.0.0 all resolved up to 2.1.20, which is how you end up with a newer stdlib under an older compiler. ~~Removing it makes the stdlib follow the compiler by itself, which is what the reflect alignment needed in the first place.~~ **Correction: that was wrong. The stdlib follows the highest requester in the graph, not the compiler — on RN <=0.73 it lands on 2.0.0 while reflect stays at 1.8.0. Matrix and full explanation here: https://github.com/klaviyo/klaviyo-react-native-sdk/pull/412#issuecomment-5585702776**

`kotlin-reflect` still tracks `$kotlin_version`, since that part was a genuine latent runtime bug.

**Why CI never caught any of this.** `example/android/build.gradle` sets `ext.kotlinVersion`, so our own build always takes the `ext` branch and **never once exercises the fallback**. Every green check on this PR was blind to the path that was broken. That is the more useful finding here, and it is why I am doing the matrix rather than just asserting the fix.

**Still to do:** pack the tarball into stock apps at 0.70, 0.72, 0.73, 0.81 and 0.86, recording loaded AGP, loaded Kotlin plugin, resolved stdlib and reflect, and the debug build result for each. Scripted rather than manual, so it can become the React Native version axis you raised earlier — which overlaps your own MAGE-1075 on `pack-and-test.sh`. I will post the table here when it runs.

Atharva8168:
Correcting my own reply above on point 3, because it read as more complete than it was.

I wrote "Splitting the Kotlin version — agreed" and then listed two fixes: the fallback reverted to `1.8.0`, and the explicit `kotlin-stdlib` removed. **Neither of those is the split you asked for.** There was still a single `KlaviyoReactNativeSdk_kotlinVersion` doing both jobs, so I had moved the incompatibility rather than separating it — the standalone `android/gradlew` was left resolving Kotlin 1.8.0 against its own Gradle 9.3.1 wrapper. CodeRabbit flagged the same thing on `android/build.gradle:25` a couple of minutes after I pushed.

Now actually done in `2577bff`, selected on `project == rootProject` in both the buildscript classpath and the dependency block, so `kotlin-reflect` follows whichever Kotlin version is selected:

| Build | Kotlin version | From |
|---|---|---|
| standalone `android/gradlew` | **2.1.20** | `KlaviyoReactNativeSdk_standaloneKotlinVersion` (new) |
| host that sets `ext.kotlinVersion` | host's | `rootProject.ext.kotlinVersion` |
| host that sets none (RN ≤ 0.72) | **1.8.0** | `KlaviyoReactNativeSdk_kotlinVersion` |

On your 2.1.20 for standalone — it matches upstream rather than being my pick. `@react-native/gradle-plugin@0.86.x` pins `kotlin = "2.1.20"` in its `libs.versions.toml` **and** ships a `gradle-9.3.1` wrapper, so React Native runs this exact pairing itself. Kotlin's published matrix stops at Gradle 8.11 for 2.1.x, which is a reason to keep the value confined to a build no consumer ever runs, but not a reason to avoid it.

Verified by resolution rather than assertion. Standalone `buildEnvironment` now loads `kotlin-gradle-plugin:2.1.20`, and `dependencies --configuration debugCompileClasspath` resolves both `kotlin-reflect` and `kotlin-stdlib` to 2.1.20 — the stdlib arriving from the plugin, since we declare none. ~~That incidentally confirms the stdlib removal was safe.~~ **Correction: that holds for the standalone build only. On RN <=0.73 the stdlib resolves to 2.0.0 while reflect stays at 1.8.0, so the removal is not safe there. See https://github.com/klaviyo/klaviyo-react-native-sdk/pull/412#issuecomment-5585702776**

**One thing I found and deliberately did not fix.** The standalone build still fails, at `:compileDebugKotlin`:

```
'compileDebugJavaWithJavac' task (current target is 1.8) and 'compileDebugKotlin'
task (current target is 17) jvm target compatibility should be set to the same
Java version.
```

`compileOptions` declares Java 1.8 and nothing sets Kotlin's target, so Kotlin takes the JDK's. I checked `origin/master` in a clean worktree and **it fails identically** — same error, with Gradle 8.14.3 and Kotlin 1.8.0. So the standalone build has never worked and this branch is not regressing it. I left it alone because aligning those targets changes the bytecode level our Kotlin compiles to inside consumer builds, which is exactly the kind of thing your merge bar rules out of this PR. Happy to ticket it — it also overlaps the standalone-`android/`-build idea from your earlier comment.

Still outstanding from your review: the packed-tarball matrix across 0.70 / 0.72 / 0.73 / 0.81 / 0.86. Worth saying plainly that the `1.8.0` fallback row above is **still unexercised** — `example/android/build.gradle` sets `ext.kotlinVersion`, so our own build always takes the host branch. The matrix is the only thing that tests it, and I will post the table here when it runs.

glenn-klaviyo:
currently doing an iOS review of this PR

glenn-klaviyo:
it would be good to rebase your `feat/react-native-86` branch and subsequently this one once the 2.5.1 patch goes out so it has the latest Podfile.lock but i tested this locally on iOS and didn't run into any hiccups. 

Screenshot 2026-09-04 at 4 20 49 PM

Atharva8168:
Thanks for testing iOS. Done in `ab1cb1f` — `feat/react-native-86` fast-forwarded to master and master merged in here, so both now carry the 2.5.1 `Podfile.lock` with klaviyo-swift-sdk 5.4.1.

Merged rather than rebased, so your four commits keep their authorship. `Podfile.lock` conflicted, so I regenerated it with `pod install` and confirmed `React-Core (0.86.3)` and the `RCT_USE_PREBUILT_RNCORE = '0'` opt-out both survived.

Atharva8168:
**Correction on the stdlib removal, with the matrix behind it.**

I claimed that removing our explicit `kotlin-stdlib` would let the stdlib follow the compiler, keeping it aligned with `kotlin-reflect`. The matrix says otherwise. Packed tarball installed into stock apps, one per version:

| RN | AGP (host declares) | AGP (we declare) | Kotlin plugin | kotlin-stdlib | kotlin-reflect | debug build |
|---|---|---|---|---|---|---|
| 0.70.15 | 7.2.1 | none (guarded) | 1.8.0 | **2.0.0** | **1.8.0** | FAIL |
| 0.72.17 | 7.4.2 | none (guarded) | 1.8.0 | **2.0.0** | **1.8.0** | pass |
| 0.73.11 | 8.1.1 | none (guarded) | 1.8.0 | **2.0.0** | **1.8.0** | pass |
| 0.81.5 | 8.11.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass |
| 0.86.3 | 8.12.0 | none (guarded) | 2.1.20 | 2.1.20 | 2.1.20 | pass |

**Your parent-first point is confirmed on all five rows.** Host AGP ranges from 7.2.1 to 8.12.0 while we declare 8.10.1, and ours never loads. The `none (guarded)` column is our own declaration being correctly inert inside a host build.

**But reflect and stdlib diverge on the three oldest rows**, and the promoter is our own dependency chain:

```
klaviyo-android-sdk:core / analytics / location  4.5.1
  └─ kotlinx-coroutines-core 1.9.0
       └─ kotlin-stdlib 2.0.0
```

Every klaviyo module requests `kotlin-stdlib-jdk8:1.9.25`; coroutines takes the base stdlib to 2.0.0. Reflect is pinned at whatever we select and does not move with it.

So the effective Kotlin floor for consumers is already **2.0.0**, and our `1.8.0` fallback sits below the stdlib every consumer actually receives. Against the compatibility table, 2.0.0 is the only value that fits all three versions that use the fallback — 2.1.20 needs Gradle >= 7.6.3 and 0.70 ships 7.5.1, while 1.9.25 caps at Gradle 8.1.1 and 0.73 ships 8.3.

**Not a regression.** I packed master into the same RN 0.70 app in a clean worktree: it resolves reflect 1.8.21 against stdlib 2.0.0, and fails to build with the same four errors. Both problems predate this branch. But they are not fixed below 0.81, and I said they were.

Raising the fallback to 2.0.0 would fix 0.70 and 0.72. It will not fix 0.73, which sets `ext.kotlinVersion = "1.8.0"` in its own template and wins parent-first. Fixing that needs reflect to track the resolved stdlib rather than our chosen version — which may be moot depending on how MAGE-1176 lands, since if we do not support 0.72 and older the fallback is dead code.

Two other things the run turned up, both pre-existing and both identical on master:

- **RN 0.70 does not build at all** — `Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8`, from `klaviyo-android-sdk` 4.5.1. Our README claims support for 0.70+.
- The standalone `android/gradlew` build fails on the same class of jvm-target mismatch, which is why two review findings were able to sit in that `buildscript` block unexercised.

I have corrected the two comments above that asserted the alignment was fixed.

android/consumer-rules.pro
```diff
@@ -0,0 +1,29 @@
...
+# below ship AGP 8.12.0 or older and do not hit it.
+#
+# This rule lives in the wrapper only to unblock React Native consumers. Native Android
+# consumers of klaviyo-android-sdk on AGP 9 hit the identical crash and are NOT covered
+# by it. The durable fixes are a consumer rule in klaviyo-android-sdk:analytics, and
+# dropping the strict pin so apps can resolve a WorkManager that ships current rules.
+# Remove this file once the Android SDK carries the rule.
+#
+# Tracked: MAGE-1199 (klaviyo-android-sdk)
+-keep class androidx.work.impl.WorkDatabase_Impl { <init>(); }
```
evan-masseau, line 36: Uh this one puzzles me a bit, you sure this is our responsibility to fix? Even in the Android SDK

android/consumer-rules.pro
```diff
@@ -0,0 +1,29 @@
...
+#     at androidx.work.WorkManagerInitializer
+#
+# androidx.work arrives here transitively:
+#   klaviyo-android-sdk:analytics -> androidx.work:work-runtime-ktx:{require 2.9.0; reject _}
+# work-runtime 2.9.0 predates AGP 9, and its own proguard.txt keeps Workers and
+# ListenableWorker constructors but says nothing about the Room database.
+#
+# Reproduced on React Native 0.87.1 (AGP 9.2.1, Gradle 9.4.1) with minifyEnabled true,
+# against both this branch and the published klaviyo-react-native-sdk@2.5.0 -- so it is
+# not specific to any wrapper change. Debug builds are unaffected. React Native 0.86 and
```
evan-masseau: If it's not specific to any wrapper code or Android SDK code, then the host app should probably decide if they want to do this...

android/consumer-rules.pro
Atharva8168: I agree with you that the bug isn't ours. But, that rule keeps the class and says nothing about its members. 
So R8 is free to drop the no-arg constructor, which is why the crash is "InstantiationException" rather than "ClassNotFoundException". 
AGP 8's R8 left it alone, AGP 9's is not. Nothing of ours is involved here.

So I would frame the keep rule not as fixing AndroidX's bug but as not shipping a launch crash with a removal condition in the file. 
Open to alternatives though, relaxing the pin is the obvious one, but I can't currently show it works.

android/build.gradle
```diff
@@ -1,15 +1,44 @@
...
+  //
+  //   - Inside a host app. Plugins load parent-first, so a host declaring its own Kotlin
+  //     plugin wins and neither value below applies. Only hosts declaring none reach the
+  //     fallback, and those are React Native 0.72 and older, whose templates ship Gradle
+  //     7.5.1 and set no ext.kotlinVersion. That value must stay low enough for 7.5.1,
+  //     which rules out 2.1.20 (it needs Gradle >= 7.6.3).
+  //
+  // Collapsing these into one property is what broke old consumers: the version the
+  // standalone build needs is far above the version an RN 0.70 host can accept.
+  def kotlin_version = project == rootProject
```
evan-masseau: nit: this syntax is kinda convoluted to put all inline, can you break it out of ternaries?

evan-masseau: switch/case or just conditional, not sure what groovy supports exactly.

android/build.gradle
```diff
@@ -1,15 +1,44 @@
...
+    // entry is never used. Declaring it unconditionally only resolves a version the host discards.
+    if (project == rootProject) {
+      classpath "com.android.tools.build:gradle:8.10.1"
+    }
+
+    // The Kotlin plugin, by contrast, must stay unconditional. React Native templates only began
+    // declaring kotlin-gradle-plugin at 0.73, and a subproject's buildscript classpath is not
+    // shared with its siblings, so on 0.72 and older this entry is the only thing that supplies
+    // `kotlin-android` to this module. A host that declares its own Kotlin plugin still wins,
+    // parent-first; kotlin_version above only takes effect when the host declares none.
```
evan-masseau, line 49: But we're going to do this PR too right? that accommodates AGP 9?
https://github.com/klaviyo/klaviyo-react-native-sdk/pull/416/files

Atharva8168: Yeah, we should take it — it's not covered by this PR.

Different lines: we guard the buildscript classpath, #416 guards `apply plugin: "kotlin-android"` at line 65, which we never touch. I tried it on an RN 0.87.1 scaffold with `android.builtInKotlin=true` — our code as-is dies with `Cannot add extension with name 'kotlin'`, add his 7 lines and it configures fine.

Not urgent though. RN 0.87 ships `builtInKotlin=false` and `newDsl=false` together, so you can't hit it from a stock app yet. That also makes CodeRabbit's "Major" on #416 a non-issue — it's warning about a flag combo the template never ships.

The red CI there isn't his change either, the branch is 6 commits behind and predates the swift-sdk 5.4.1 pin. Android all passes, the failure is an Xcode plist thing.

I'd land this first, then update his branch — needs a rebase anyway since we rewrote that region.

android/build.gradle
Atharva8168: Done — went with `if / else if / else` and a comment per branch. Groovy does have `switch`, but the conditions aren't all on the same value (one's `project == rootProject`, one's `ext.has(...)`), so it'd be `switch (true)` which reads worse than the if-chain.

No behaviour change, standalone still resolves kotlin-gradle-plugin 2.1.20.

evan-masseau (approved):

evan-masseau:
appreciate the thoroughness in your approach on this, and your patience with all my nitpicking.

Atharva8168:
Apologies for the overly verbose comments. I'll work on keeping them shorter and better curated in future.

PR #414, opened 2026-09-03
chore(release): merge rel/2.5.1 into master [MAGE-1183]

Atharva8168:
## Summary

Merge-back for the **2.5.1** release. Tracked in MAGE-1183.

**Do not merge until 2.5.1 is tagged and published.** In this repo the release order is: land the work on `rel/2.5.1` → tag the branch tip → publish a GitHub Release (that is what pushes to npm) → *then* merge this PR. Merging early puts the version bump on `master` while npm still has 2.5.0.

## Current contents

Only the version bump so far (`5203d86`). The actual cascade — raising the pinned native SDK versions for the MAGE-1177 forms session-timeout fix — is still to come via PR #413, and will appear here once it lands on `rel/2.5.1`.

## Precedent

The 2.5.0 merge-back was a squash onto `master` (`a9a68fe` "rel/2.5.0 (#403)"). Worth matching so `master`'s history stays one-commit-per-release.

Skipping the merge-back is how `rel/2.4.2` ended up orphaned on origin with `package.json` still at `2.4.1` and no tag.

Refs MAGE-1183, MAGE-1181, MAGE-1180

PR #411, opened 2026-08-27
chore(deps): consolidate protobufjs, ip-address and js-yaml npm bumps

evan-masseau:
# Description

Consolidates three pending Dependabot npm bumps into a single PR — one CI run and one review instead of three. Supersedes #385, #395, and #399.

## Due Diligence

- [x] I have tested this on a simulator/emulator or a physical device, on iOS and Android (if applicable).
  - N/A — transitive dev-dependency bumps with no runtime surface. Verified via the full CI gate set locally (see Test Plan).
- [x] I have added sufficient unit/integration tests of my changes.
  - N/A — dependency version bumps, no behavior change. Existing suite (79 tests) passes unchanged.
- [x] I have adjusted or added new test cases to team test docs, if applicable.
  - N/A
- [x] I am confident these changes are implemented with feature parity across iOS and Android (if applicable).
  - N/A — JS toolchain deps only, no native code touched.

## Release/Versioning Considerations

- [x] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [ ] `Minor` Contains changes to the public API.
- [ ] `Major` Contains **breaking** changes.
- [ ] Contains readme or migration guide changes.
- [ ] This is planned work for an upcoming release.
  - Dependency hygiene, not planned feature work. No release plan impact — none of these packages ship in the published artifact; all three are build/dev-time transitives.

## Changelog / Code Overview

Each bump is the exact version Dependabot proposed, cherry-picked from the Dependabot commits rather than hand-edited, so the resolved versions and checksums are provably unmodified.

| Supersedes | Package | From → To | Why |
|---|---|---|---|
| #385 | `protobufjs` | 7.6.4 → 7.6.5 | Bug fix: handle EOF during options parsing (protobufjs/protobuf.js#2356) |
| #395 | `ip-address` | 10.2.0 → 10.4.0 | Byte-array validation for `Address6`, node 12 loadability |
| #399 | `js-yaml` | 3.15.0 → 3.15.1 | **Security**: removes quadratic complexity from `!!omap` duplicate-key detection |

Total diff: 4 changed lines across 2 files.

### One `package.json` edit is required, by design

[...]

PR #415, opened 2026-09-03
fix(ios): cascade the forms session-timeout fix — pin klaviyo-swift-sdk 5.4.1 [MAGE-1181]

Atharva8168:
# Description

Cascades the in-app forms session-expiry fix from `klaviyo-swift-sdk` 5.4.1 into the React Native SDK for the **2.5.1** patch release. This is a native pin bump only — no TypeScript or bridge changes are required.

## Due Diligence

**Testing status — please read before approving.** The device smoke test has **not** been run yet. The behaviour this release repairs is only observable by expiring a forms session on a device, so CI passing does not prove the cascade worked. See the Test Plan below.

No new tests were added because no first-party code changed. Android parity already holds: upstream swift PR #715 states the fix brings iOS to parity with `klaviyo-android-sdk`, which is why the Android pin does not move.

Patch is the correct level. Upstream PR #715 is explicit that its changes are internal to `IAFPresentationManager` with no public API changes, and nothing in this repo's public surface moves.

## Changelog / Code Overview

Two files:

| File | Change |
| --- | --- |
| `klaviyo-react-native-sdk.podspec` | `KlaviyoSwift`, `KlaviyoLocation`, `KlaviyoForms` pinned `5.4.0` → `5.4.1` |
| `example/ios/Podfile.lock` | regenerated via `pod update`; all five Klaviyo pods now resolve at `5.4.1` |

### Why no wrapper code changes

The podspec pins exact versions, so the bump has to be explicit. Beyond that, the fix reaches consumers through the pinned pod alone:

- The only session-timeout surface is `sessionTimeoutDuration` (`src/Forms.ts:33`), which both bridges pass straight through to the native `InAppFormsConfig` initializer — `ios/KlaviyoBridge.swift:89-90` and `KlaviyoReactNativeSdkModule.kt:146-149`.
- Upstream PR #715 changes only `IAFPresentationManager` internals and does not alter `InAppFormsConfig`'s public shape, so none of those three files need to move.

### What the upstream fix repairs

When a form session expired, the old expiry path destroyed only the webview. On the rebuilt webview's handshake, `startProfileObservation()`'s idempotency guard saw a still-live observer and short-circuited, so the fresh webview never re-subscribed to the event bus and trigger-based forms silently stopped firing. 5.4.1 adds a scoped `tearDownFormWebView()` that clears the observer and its tasks first, plus two adjacent lifecycle repairs.

### Deliberately not changed

- `android/gradle.properties` stays at `4.5.1`. The bug was iOS only, and `4.5.1` is the newest Android release.
- `package.json`, the iOS plist, Android `strings.xml`, and the example app version strings are already at `2.5.1` from `5203d86` on `rel/2.5.1`. They are not re-bumped here.

### One incidental line

`example/ios/Podfile.lock` also shows an `RCT-Folly` checksum change with its version unchanged at `2024.11.18.00`. React Native is `0.81.5` in both `example/package.json` and `node_modules`, so this is not a version skew — the previously committed lock was generated against a slightly different local podspec state. Flagging it rather than hand-editing one checksum back, which would leave the lock inconsistent with what `pod install` actually computes.

## Test Plan

**Not yet executed.** Reviewers should treat this as outstanding.

1. Lower `sessionTimeoutDuration` in `example/src/hooks/useForms.ts` from `3600` to about `30`.
2. Run the example app on an iOS simulator and register for in-app forms.
3. Configure a form that triggers on a tracked event.
4. Background the app for longer than the timeout, then foreground it.
5. Fire the trigger event and confirm the form appears.
6. Repeat the background/foreground cycle a second time and confirm the form still appears.
7. Revert the `sessionTimeoutDuration` edit.

Run the same steps once on Android as a regression check — the Android pin does not move, so the result must be unchanged. Then verify a third-party install resolves the new pod with `./pack-and-test.sh setup` followed by an example app build.

## Related Issues/Tickets

- MAGE-1181 — this ticket, the React Native half of the cascade
- MAGE-1180 — parent, cascade to both wrappers
- MAGE-1177 — the origin iOS fix
- MAGE-1183 — 2.5.1 release mechanics
- Upstream: https://github.com/klaviyo/klaviyo-swift-sdk/pull/715

## Summary by CodeRabbit

* **Chores**
  * Updated supporting Klaviyo components to version 5.4.1.
  * Existing optional component controls remain unchanged.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -63,21 +63,21 @@ PODS:
   - hermes-engine (0.81.5):
     - hermes-engine/Pre-built (= 0.81.5)
   - hermes-engine/Pre-built (0.81.5)
-  - klaviyo-react-native-sdk (2.5.0):
-    - KlaviyoForms (= 5.4.0)
-    - KlaviyoLocation (= 5.4.0)
-    - KlaviyoSwift (= 5.4.0)
+  - klaviyo-react-native-sdk (2.5.1):
+    - KlaviyoForms (= 5.4.1)
+    - KlaviyoLocation (= 5.4.1)
+    - KlaviyoSwift (= 5.4.1)
     - React-Core
-  - KlaviyoCore (5.4.0):
+  - KlaviyoCore (5.4.1):
     - AnyCodable-FlightSchool
-  - KlaviyoForms (5.4.0):
-    - KlaviyoCore (~> 5.4.0)
-  - KlaviyoLocation (5.4.0):
-    - KlaviyoSwift (~> 5.4.0)
-  - KlaviyoSwift (5.4.0):
+  - KlaviyoForms (5.4.1):
+    - KlaviyoCore (~> 5.4.1)
+  - KlaviyoLocation (5.4.1):
+    - KlaviyoSwift (~> 5.4.1)
+  - KlaviyoSwift (5.4.1):
     - AnyCodable-FlightSchool
-    - KlaviyoCore (~> 5.4.0)
-  - KlaviyoSwiftExtension (5.4.0)
+    - KlaviyoCore (~> 5.4.1)
+  - KlaviyoSwiftExtension (5.4.1)
   - nanopb (3.30910.0):
     - nanopb/decode (= 3.30910.0)
     - nanopb/encode (= 3.30910.0)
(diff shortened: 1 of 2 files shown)


glenn-klaviyo (approved):
LGTM

belleklaviyo:
can you also update the marketing version in the example ios app? see [here](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/403/changes#diff-b6ae90b73703cf5f8f165bcd8f0a63993dfbd2f7ae5d8430fb7696e5ce74aeeb)

glenn-klaviyo:
i thought this ^ too but i saw it was bumped in [https://github.com/klaviyo/klaviyo-react-native-sdk/pull/414/changes](<https://github.com/klaviyo/klaviyo-react-native-sdk/pull/414/changes>)

Atharva8168:
@belleklaviyo good catch, and it is already done — just one commit earlier than the #403 precedent would lead you to expect.

The marketing version was bumped in [`5203d86`](https://github.com/klaviyo/klaviyo-react-native-sdk/commit/5203d86) ("chore(release): bump version to 2.5.1"), which already sits on `rel/2.5.1` — the base branch of this PR. All four `MARKETING_VERSION` entries in `example/ios/KlaviyoReactNativeSdkExample.xcodeproj/project.pbxproj` read `2.5.1` on this branch today.

That commit covered the full set, not just the pbxproj:

| File | Field |
| --- | --- |
| `package.json` | `version` |
| `ios/klaviyo-sdk-configuration.plist` | `klaviyo_sdk_version` |
| `android/src/main/res/values/strings.xml` | `klaviyo_sdk_version_override` |
| `example/android/app/build.gradle` | `versionName` |
| `example/ios/.../project.pbxproj` | `MARKETING_VERSION` |

**Why it looks missing here.** For 2.5.0 the bump lived inside the release PR, so it appeared in that diff. For 2.5.1 it was split into its own commit on `rel/2.5.1` *before* the cascade work, so this PR is scoped to the native pin bump only and its diff correctly shows nothing. The base already carries it.

@glenn-klaviyo you're right that #414 shows it — that PR is `rel/2.5.1 → master` and carries `5203d86`. Pointing at `5203d86` on `rel/2.5.1` directly is probably the less confusing reference, since #414 is the merge-back rather than where the bump originates.

belleklaviyo:
ohhh ok cool cool all good then

Atharva8168:
Superseded by #419.

This PR merged into the old `rel/2.5.1`, which was branched from the `master` tip and so inherited six Dependabot commits (#391, #392, #411). A patch release should carry only its own fix, so the release branch was re-cut.

The work here was re-applied onto **`rel/2.5.1-patch`**, cut from `a9a68fe` — the 2.5.0 point on `master`. Same file changes, new commit IDs. See #419.

Nothing is lost: the old branch tip is preserved at `backup/rel-2.5.1-old` (`0ad1287`), and this PR stays merged for the record.

PR #423, opened 2026-09-10
chore(scripts): add React Native version matrix tool

Atharva8168:
# Description

Adds `scripts/version-matrix/`, a manual tool that packs this working tree exactly as npm would publish it, installs that tarball into stock React Native apps at several versions, and records what actually happens. It supersedes the ad-hoc scripts used during MAGE-919.

## Due Diligence

  - **All four tiers, all 12 React Native versions, 12/12 clean.** Full matrix in the Test Plan. Tier 4 means a packed-tarball install, a minified release build, then install, launch and read logcat on an emulator.
  - This *is* test tooling. `smoke/App.tsx` is typechecked against the real SDK API by the existing `yarn typecheck`, because `tsconfig.json` maps `klaviyo-react-native-sdk` to `./src/index`.
  - Android only, deliberately. See "Not covered" in the README.

  - Developer tooling, no runtime impact. Nothing here ships to npm **as of `74da317`** — but see the note below, because that was not true when this PR was opened.

### Why

CI builds exactly one thing: `example/` at one React Native version, debug only. Two properties of that setup hide whole classes of bug.

- `example/android/build.gradle` sets `ext.kotlinVersion`, so every CI build takes the host branch of our Kotlin version selection. The fallback older consumers get is never executed.
- `example/` resolves the SDK through a yarn-workspace symlink, not the `files[]` allowlist a customer installs. Those are different file sets.

So CI cannot see version-selection bugs, packaging bugs, consumer-side AAR metadata rejections, or anything that only appears in a minified release build. All four bit us during MAGE-919.

### What

| File | |
|---|---|
| `version-matrix.sh` | The tool. Four tiers, each a superset of the last. |
| `smoke/App.tsx` | The tier-4 API smoke app, copied over the scaffold before the release build. |
| `README.md` | Usage, how to read each column, and what every guard prevents. |

| Tier | Adds |
|---|---|
| 1 | Resolution facts only — AGP, Kotlin plugin, stdlib, reflect. No compilation. |
| 2 | + debug build of our module, then of the consumer app (runs AAR metadata checks) |
| 3 | + release build with R8, recording which ProGuard config actually applied |
| 4 | + install, launch, exercise the public API, read logcat |

Outputs `matrix.md`, `results.json` (for diffing runs before and after a change), and per-version raw logs.

### Seven defects from MAGE-919 that this fixes

Each of these produced a wrong reading during that investigation.

| Guard | What went wrong without it |
|---|---|
| Scaffolds snapshotted pristine, restored before **every** run | A cached scaffold still carried `kotlinVersion = "2.0.0"` from an earlier experiment and produced a false stdlib/reflect reading. Caught by luck. An existence check cannot detect a mutated scaffold. |
| Tarball filename carries a hash of its **contents** | npm caches `file:` dependencies by path. A repack under the same name was silently ignored and a stale tarball installed. |
| `diff -r` installed package vs packed tarball | Nothing checked this. A fix was once "verified" against a package that did not contain the fix. |
| `rm -rf app/build/outputs` before each release build | A build failed in 3s and the script installed the *previous* APK. Voided a whole matrix cell. |
| Smoke calls `isLoggingEnabled` | It is the one public API taking a callback. Every other call is fire-and-forget on the legacy bridge, so a JS `try/catch` proves nothing. "108 calls, zero failures" was measuring nothing. |
| R8 confirmed by `mapping.txt` existing | Verifies the outcome, not that a `sed` matched. |
| ProGuard config recorded | `proguard-android.txt` contains `-dontoptimize`. 11 of 12 "minified" runs during MAGE-919 silently used it. |

Also: pure ASCII and bash 3.2 idioms throughout (macOS folds a multibyte character following `$var` into the variable *name*, and `${VAR,,}` does not exist there), and the workspace lives outside `/tmp`, which macOS purges.

### It found a real bug on its first run

Tier 1 against `master` reports a Kotlin version skew:

```
| RN     | AGP (host) | AGP (ours) | Kotlin plugin | kotlin-stdlib | kotlin-reflect | skew |
| 0.86.3 | 8.12.0     | 8.10.1     | 2.1.20        | 2.1.20        | 1.8.21         | yes  |
```

`android/build.gradle:147` on master hardcodes `kotlin-reflect:1.8.21` while `kotlin-stdlib` resolves to `2.1.20` from the RN 0.86 template. The tool exits non-zero on this, so it works as a pre-release gate.

PR #412 already changes that line to track `$kotlin_version`. MAGE-1203 removes the dependency entirely.

## Test Plan

Verified on this branch:

```bash
./scripts/version-matrix/version-matrix.sh --tier 1 0.86.3
```

- Pack produced a content-hashed tarball, 71 files
- Scaffold created and snapshotted (836K, `node_modules` excluded)
- `install-verify.log` is 0 bytes — installed package matches the tarball exactly
- Skew detected, reported in table and JSON, exit code 1
- `results.json` parses
- `yarn typecheck` and `yarn lint` clean

### All four tiers, 12 versions, on the fixed tool

Run at tier 4 against `asb/mage-1203-drop-kotlin-reflect` @ `a290684` (PR #424), one identical tarball across every row.

| RN | AGP host | AGP ours | KGP | stdlib | reflect | skew | mod dbg | app dbg | proguard | R8 | release | crash | JS smoke | rt | alive | constants |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 0.76.9 | 8.6.0 | none | 1.9.25 | 2.0.0 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.77.3 | 8.7.2 | none | 2.0.21 | 2.0.21 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.78.3 | 8.8.0 | none | 2.0.21 | 2.0.21 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.79.7 | 8.8.2 | none | 2.0.21 | 2.0.21 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.80.3 | 8.9.2 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.81.5 | 8.11.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.82.1 | 8.12.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.83.10 | 8.12.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.84.1 | 8.12.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.85.3 | 8.12.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.86.3 | 8.12.0 | none | 2.1.20 | 2.1.20 | absent | - | pass | pass | proguard-android.txt | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |
| 0.87.1 | 9.2.1 | none | 2.2.0 | 2.2.10 | absent | - | pass | pass | **proguard-android-optimize.txt** | yes | pass | 0 | 10/10 | yes | yes | 6de39d683b4c |

`install verified against tarball` and a zero-byte `diff -r` on every version. Zero tool defects across the ~15 changes in `2d94999` and `74da317` — that sweep doubled as the only real test of them.

Caveat kept rather than buried: 11 of 12 rows used `proguard-android.txt`, which carries `-dontoptimize`. R8 ran everywhere (`mapping.txt` present), but the optimisation pass only ran on 0.87.1.

## What review changed, and one thing this PR got wrong about itself

Four independent model reviews plus two `mage:reviewer` passes went over this. They found real defects, several of them in the guards themselves. Worth reading if you review tooling here, because the failure shape repeats.

**This PR shipped the tool to npm, and said in writing that it didn't.** The original body claimed `scripts/` is absent from `files[]` so nothing ships. True, and irrelevant: `lib` **is** in `files[]`, and `bob` compiled the smoke app into `lib/typescript/scripts/`:

```
npm notice 105B lib/typescript/scripts/version-matrix/smoke/App.d.ts
npm notice 175B lib/typescript/scripts/version-matrix/smoke/App.d.ts.map
```

A PR whose whole thesis is that CI cannot see packaging changes introduced one. The tool could not catch it either — the files were in both the tarball and the install, so `diff -r` passed. **Three of four reviewers repeated the same wrong conclusion**, having checked `files[]` and stopped. Fixed by excluding `scripts` in `tsconfig.build.json`; back to 69 files.

**Five paths could report a pass that meant nothing.** Three `gradlew` exit codes were discarded, so a Gradle that never ran gave `?` cells and exit 0 — on the tier the README calls the default. `none (guarded)` was indistinguishable from a crashed build while the README called that absence proof the guard works. `FATAL EXCEPTION` was counted and ignored. `pidof` computed the right answer and threw it into stderr. And the constants probe added in `645b223` was read by nothing.

**The parser missed the form Gradle actually emits.** `resolved_version` returned nothing for `kotlin-stdlib:{strictly 2.1.20} -> 2.1.20`, which is in this repo's own logs. It worked only because plain-form lines coexisted. Also fixed: version-less constraints, dynamic `+`, reading a version off a `FAILED` line, and truncating `2.2.0-RC2` to `2.2.0` — which made a real skew compare equal.

**A fix commit made one case worse.** `2d94999` replaced the `logcat -c` reliance with a timestamp filter and applied it to the authoritative read but not the wait-loop probe. On a device where the clear fails, that turns a false pass into a false *failure*. Caught by the second review round, fixed in `74da317`.

**The README's own premise was wrong.** It claimed CI builds one thing, debug only. `publish-example.yml` runs `pack-and-test.sh` then `:app:bundleRelease` on master. The gap this tool actually closes is the version axis and minification — packaging fidelity was already covered.

A `bash -n` gate now runs in the lint job. It catches parse errors only; bash 3.2 semantics still parse cleanly, so it is not a compatibility check.

## Related Issues/Tickets

- Not tied to a ticket. Built while working on the React Native 0.81 to 0.86 upgrade, which is where the seven defects above were found, but this tool stands on its own and ships independently.
- MAGE-1203 — removes `kotlin-reflect`, which deletes the `reflect` column and both skew findings
- MAGE-1176 — which React Native versions we support. Blocks any future CI version axis, and should drive `DEFAULT_VERSIONS`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Atharva8168:
Closing this. The version matrix tool is moving to `klaviyo-react-native-test-suite`, which is private. It's internal test tooling and doesn't need to sit in a public repo.

The script, smoke app and README all move over, plus the fixes from #427.

The only thing that doesn't travel is the bash -n check this PR added to ci.yml, since there'll be no script here to lint.

PR #426, opened 2026-09-18
feat(android): React Native 0.86 support (rel/2.6.0) [Part of MAGE-919]

Atharva8168:
Merges the React Native 0.86 integration branch into `rel/2.6.0`.

Part of MAGE-919.

## What lands

Four commits, all already reviewed and approved on their own PRs:

| PR | Change |
| --- | --- |
| #412 | React Native 0.81.5 -> 0.86.3, and a rewrite of the Kotlin version selection in `android/build.gradle` |
| #416 | Skip `apply plugin: "kotlin-android"` when AGP 9 has already registered the `kotlin` extension (external contribution from @gabrieldonadel) |
| #424 | Drop `kotlin-reflect` from `getConstants()`, replacing reflection with two static maps (Closes MAGE-1203) |

`android/consumer-rules.pro` is also new — a `WorkDatabase_Impl` keep rule that stops an AGP 9 launch crash. It is a documented stopgap and names its own removal condition; see MAGE-1199 and MAGE-1213.

## Why this is a minor, not a major

The published contract does not change.

- `peerDependencies` are byte-identical to `master`. The npm install contract is untouched.
- The AGP classpath entry is guarded with `if (project == rootProject)`, so a host app's own AGP wins parent-first. We do not push AGP 9 onto anyone.
- #416's guard is a no-op on AGP 8 — `findByName('kotlin')` reads null there and the plugin applies exactly as before.
- The Gradle wrapper bumps are confined to `android/` and `example/android/`, neither of which a consumer builds.

## Verification

A version matrix packed this tree as npm would publish it and installed that tarball into stock React Native apps at 12 versions:

```
0.76.9  0.77.3  0.78.3  0.79.7  0.80.3  0.81.5
0.82.1  0.83.10 0.84.1  0.85.3  0.86.3  0.87.1
```

All 12 passed at tier 4 — installed, launched, exercised the public API, and round-tripped the bridge. `kotlin-reflect` reads absent on every row, and the `getConstants()` digest is identical across all 12, confirming the reflection removal changed nothing on the wire.

## Known gaps

- **Nothing below 0.76.9 was tested.** The README still claims 0.70+, and the Kotlin 1.8.0 fallback for RN 0.72-and-older has never been exercised in CI or in the matrix. MAGE-1176 tracks settling the supported range, and this release note cannot state one until it does.
- **Physical-device push testing is outstanding** on real iOS and Android hardware. It is in MAGE-919's acceptance criteria.
- **`kotlin-reflect` left the runtime classpath.** It was `implementation`, never `api`, so it was never on a consumer's compile classpath. A consumer resolving it transitively through us and using it at runtime would lose it.

## Still to do on this release branch

- `chore(release)` PR bumping `package.json` to 2.6.0 (it currently reads 2.5.1)
- Release notes

## Summary by CodeRabbit

* **Compatibility**
  * Updated support for React Native 0.86.3, React 19.2.8, Node.js 22+, Kotlin, and Gradle 9.3.1.
  * Updated the example app and development tooling to align with the latest React Native release and new architecture.

* **Bug Fixes**
  * Improved Android release builds by preventing reflective WorkManager initialization failures.
  * Made SDK constants more reliable and consistent across supported versions.

* **Build Improvements**
  * Improved Android and iOS example project configuration and build compatibility.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -25,7 +25,13 @@ def isNewArchitectureEnabled() {
 }
 
 apply plugin: "com.android.library"
-apply plugin: "kotlin-android"
+// AGP 9 ships built-in Kotlin support and registers the `kotlin` extension
+// itself. Applying the Kotlin plugin on top of it fails configuration with
+// "Cannot add extension with name 'kotlin'". Only apply it when nothing has
+// registered that extension yet.
+if (project.extensions.findByName('kotlin') == null) {
+  apply plugin: "kotlin-android"
+}
 
 def getExtOrDefault(name) {
   return rootProject.ext.has(name) ? rootProject.ext.get(name) : project.properties["KlaviyoReactNativeSdk_" + name]


amber-klaviyo (approved):
Looks like the PRs have already been individually reviewed, so just auto approving the feature branch

